PR #1, opened 2015-04-07
PHPDoc fixes and simplifications.

Furgas:


stelgenhof:
@Furgas Thanks for the fixes!

PR #2, opened 2015-04-07
Add PHPUnit as require-dev in composer.json.

Furgas:


stelgenhof:
I had PHPUnit included before in my local copy, however didn't want to have PHPUnit loaded as part of Yasumi since I have it installed globally. Do you think it is still good to be included as require-dev?

Furgas:
I think that with this, it just works for every contributor, no matter their development environment. Also, you are able to point to certain PHPUnit version that works, in case they will make some backward incompatible changes.

stelgenhof:
Yes, makes sense! I will merge your PR.

PR #4, opened 2015-04-07
Added AbstractProvider::getHolidaysOfType

gisostallenberg:
this makes it possible to get only Holidays of a certain type. This makes it possible to get 'national holidays' as mentioned in the Composer description

### DIFF
--- a/src/Yasumi/Provider/AbstractProvider.php
+++ b/src/Yasumi/Provider/AbstractProvider.php
@@ -177,6 +177,22 @@ abstract class AbstractProvider implements ProviderInterface, Countable, Iterato
     }
 
     /**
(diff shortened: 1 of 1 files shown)


stelgenhof:
Thank you for this! This is certainly something missing now in Yasumi. However, I was thinking making something similar to what PEAR Date_Holidays is using: Filters. That seems to be a very flexible solution.

stelgenhof:
I just released v1.0.0 of Yasumi, which includes filters. This allows you to filter holidays to only get 'national' holidays as you requested. :)

gisostallenberg:
Thanks for adding this. Nice feature.

stelgenhof:
Glad you like it! If you have any more wishes, ideas please create an issue :). What other countries would you like to see added first?

PR #6, opened 2015-04-08
Duplicate Armistice Day in Belgium.

Furgas:


stelgenhof:
Thanks! I must have missed that one. (Copy-paste issue).

PR #9, opened 2016-04-04
Added support for Germany

eaglefsd:
Not that familiar with open source and tests, if I'm missing something, just let me know!


src/Yasumi/data/translations/ascensionDay.php
```diff
@@ -19,5 +19,6 @@
     'da_DK' => 'Kristi Himmelfartsdag',
     'nb_NO' => 'Kristi himmelfartsdag',
     'sv_SE' => 'Kristi himmelsfärds dag',
-    'fi_FI' => 'Helatorstai'
+    'fi_FI' => 'Helatorstai',
+    'de'    => 'Christi Himmelfahrt'
```
stelgenhof: Although 'de' is an official locale, it is better to use a locale that identifies also the region as well. In this case 'de_DE' is preferred. Can you adjust your PR accordingly?

tests/Germany/GermanyBaseTestCase.php
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Timezone in which this provider has holidays defined
+     */
+    const TIMEZONE = 'Europe/Berlin';
+
+    /**
+     * Locale that is considered common for this provider
+     */
+    const LOCALE = 'de';
```
stelgenhof: Although 'de' is an official locale, it is better to use a locale that identifies also the region as well. In this case 'de_DE' is preferred. Can you adjust your PR accordingly?

src/Yasumi/Provider/Germany.php
```diff
@@ -0,0 +1,69 @@
...
+        $this->timezone = 'Europe/Berlin';
+
+        // Add common holidays
+        $this->addHoliday($this->newYearsDay($this->year, $this->timezone, $this->locale));
+
+        // Add common Christian holidays (common in Germany)
+        $this->addHoliday($this->ascensionDay($this->year, $this->timezone, $this->locale));
+        $this->addHoliday($this->christmasDay($this->year, $this->timezone, $this->locale));
+        $this->addHoliday($this->easterMonday($this->year, $this->timezone, $this->locale));
+        $this->addHoliday($this->goodFriday($this->year, $this->timezone, $this->locale));
```
stelgenhof: Is Good Friday an official holiday, i.e. offices/shops are closed?

src/Yasumi/Provider/Germany.php
```diff
@@ -0,0 +1,69 @@
...
+     * the name addresses neither the re-union nor the union, but the unity of Germany. The Day of German Unity on
+     * 3 October has been the German national holiday since 1990, when the reunification was formally completed. It
+     * is a legal holiday for the Federal Republic of Germany.
+     *
+     * @link https://en.wikipedia.org/wiki/German_Unity_Day
+     */
+    public function calculateGermanUnityDay()
+    {
+        if ($this->year >= 1990) {
+            $this->addHoliday(new Holiday('germanUnityDay', ['de' => 'Tag der deutschen Einheit'],
```
stelgenhof: I believe it is written as this: "Tag der Deutschen Einheit".

stelgenhof:
Thank you very much for the PR @eaglefsd! Very much appreciated. I left a few comments on your PR. 

Also as part of the unit tests, there should also be a <COUNTRYNAME>Test.php (see for example Yasumi/tests/Finland/FinlandTest.php). This test contains tests to confirm all the holidays that you expect for this country.

In the future, I would like also to see the individual German States' holidays as they differ from the national ones; in the same way as done in Spain.

If you could review your PR based on my comments, I can merge it afterwards.

eaglefsd:
Well, I used 'de'  because you used it in the locales :) 
But I can change it, sure. 
Yes, good Friday is a day where the shops are closed, we even have a rule which forbids music in clubs. 
I didn't have the time yesterday for the individual parts, but I will definetely have a look later today!

eaglefsd:
Hi, I can commit the different states with their holidays, but I don't have the time to write the tests at the moment :/

stelgenhof:
That would be great! I can do the unit tests for you. In that case I prefer you prepare the PR for just Germany without the state holidays first. I would like to do then a next update for the German state holidays.

eaglefsd:
I removed the states in my latest commit, do I have to do anything else?
The translations for the state holidays (like epiphany) are still present tho, I hope you don't mind...

stelgenhof:
Sounds good. I will merge it and if something needs to be altered we can do that.

PR #10, opened 2016-04-04
Added support for Greece

sebdesign:


src/Yasumi/Provider/Greece.php
```diff
@@ -0,0 +1,154 @@
...
+     *
+     * @param  int    $year     the year for which Easter needs to be calculated
+     * @param  string $timezone the timezone in which Easter is celebrated
+     *
+     * @return \Datetime date of Orthodox Easter
+     *
+     * @link http://php.net/manual/en/function.easter-date.php#83794
+     * @link https://en.wikipedia.org/wiki/Computus#Adaptation_for_Western_Easter_of_Meeus.27_Julian_algorithm
+     */
+    public function calculateEaster($year, $timezone)
```
stelgenhof: Perhaps renaming this to 'calculateOrthodoxEaster' to avoid confusion with the common Christian Easter?

tests/Greece/GreeceTest.php
```diff
@@ -0,0 +1,92 @@
...
+    {
+        $this->assertDefinedHolidays([], self::REGION, $this->year, Holiday::TYPE_OTHER);
+    }
+
+    /**
+     * Initial setup of this Test Case
+     */
+    protected function setUp()
+    {
+        $this->year = $this->generateRandomYear(1945);
```
stelgenhof: The year 1945 might need to be adjusted as it might cause false errors when one of the other holidays does not exist after 1945. (This is a bit of a funny implementation in the unit tests, so I will have a look at it). Changing it will also make the CI tests succeed.

stelgenhof:
Thanks for the PR! I have left a few comments on there. Once you have adjusted the PR I will merge it. Thanks again for the contribution!

src/Yasumi/Provider/Greece.php
sebdesign: Yes that's a good point but this way I would have to override my the other holidays that depend on easter, right? E.g. Easter Monday, Pentecost, etc.

tests/Greece/GreeceTest.php
sebdesign: Ah you're right, I'll fix that.

src/Yasumi/Provider/Greece.php
stelgenhof: I haven't looked into it yet, but do Easter Monday, Pentecost then rely also on the Orthodox Easter? Let me investigate a bit further.

sebdesign: Indeed, for every holiday that relies on the common Christian Easter, in Greece it relies on the Orthodox Easter. By overriding `calculateEaster` there's no need to change the related holiday functions. If I had to use a `calculateOrthodoxEaster`, then each holiday that is calling `calculateEaster` would have to be overridden to use `calculateOrthodoxEaster`.

stelgenhof: Ah, thanks! That makes sense indeed. Let me check your PR a little bit further and then I will merge it.

sebdesign: As a sidenote, even the catholic church of Greece celebrates Easter on the same date as the Orthodox Easter in Greece, for convenience.

src/Yasumi/Provider/Greece.php
```diff
@@ -0,0 +1,154 @@
...
+        $this->addHoliday($this->pentecostMonday($this->year, $this->timezone, $this->locale));
+        $this->addHoliday($this->assumptionOfMary($this->year, $this->timezone, $this->locale));
+        $this->addHoliday($this->christmasDay($this->year, $this->timezone, $this->locale));
+
+        // Calculate other holidays
+        $this->calculateThreeHolyHierarchs();
+        $this->calculateCleanMonday();
+        $this->calculateIndependenceDay();
+        $this->calculateOhiDay();
+        $this->calculatePolytechnio();
```
stelgenhof: Are Ochi-Day and Polytechnio official national holidays/non-working days? From Wikipedia I couldn't see these as non-workingdays.

sebdesign: Ochi day is a national non-working holiday. Polytechnio as well as The Three Hierarchs are working days, but schools and universities are closed.

stelgenhof:
I cannot yet merge this PR as there are some conflicts. I guess because there are some commits done after you have created the PR. Could you check the PR? I think you can do it quicker than me :)

sebdesign:
Sure, I'll pull the latest commits and resolve the merge conflicts. :)

sebdesign:
Thank for merging the PR! (I just noticed I didn't add anything to the readme.md)

stelgenhof:
No worries. I just updated the docs :)

Issue #11, opened 2016-04-06
Norway tests failing


badams:
Some tests for the Norway provider are failing on my local machine, I've tracked it down to the timezone being set inconsistently between `Yasumi\Provider\Norway` and `Yasumi\Tests\Norway\NorwayBaseTestCase`

Output from PHPUnit:
```
PHPUnit 4.8.24 by Sebastian Bergmann and contributors.

Runtime:	PHP 5.5.30 with Xdebug 2.3.3
Configuration:	/Users/byron/projects/badams/yasumi/phpunit.xml

.............................................................   61 / 1166 (  5%)
.............................................................  122 / 1166 ( 10%)
.............................................................  183 / 1166 ( 15%)
.............................................................  244 / 1166 ( 20%)
.............................................................  305 / 1166 ( 26%)
.............................................................  366 / 1166 ( 31%)
.............................................................  427 / 1166 ( 36%)
.............................................................  488 / 1166 ( 41%)
.............................................................  549 / 1166 ( 47%)
.............................................................  610 / 1166 ( 52%)
.............................................................  671 / 1166 ( 57%)
....F...FF...F.F.............FF..F.F.F.F.F....FF...........FF  732 / 1166 ( 62%)
.F..FF.......................................................  793 / 1166 ( 68%)
.............................................................  854 / 1166 ( 73%)
.............................................................  915 / 1166 ( 78%)
.............................................................  976 / 1166 ( 83%)
............................................................. 1037 / 1166 ( 88%)
............................................................. 1098 / 1166 ( 94%)
............................................................. 1159 / 1166 ( 99%)
.......

Time: 3.5 seconds, Memory: 41.00Mb

There were 19 failures:

1) Yasumi\Tests\Norway\AscensionDayTest::testAscensionDay
Failed asserting that two DateTime objects are equal.
--- Expected
+++ Actual
@@ @@
-1877-05-10T00:00:00+0043
+1877-05-10T00:00:00+0050

/Users/byron/projects/badams/yasumi/tests/YasumiBase.php:89
/Users/byron/projects/badams/yasumi/tests/Norway/AscensionDayTest.php:35

2) Yasumi\Tests\Norway\ChristmasDayTest::testHoliday with data set #2 ('1410', DateTime Object (...))
Failed asserting that two DateTime objects are equal.
--- Expected
+++ Actual
@@ @@
-1410-12-25T00:00:00+0043
+1410-12-25T00:00:00+0050

[...]

PR #12, opened 2016-04-06
Setting Norway timezone to Europe/Oslo

badams:
Fix for #11


stelgenhof:
Thank you @badams for catching this one! Copy/paste mistake I guess. Thanks for the fix; I will merge it today.

badams:
No problem, I'm not sure why this wasn't occurring on travis, perhaps something to do with the fact that NZ just changed to daylight saving hours

Issue #15, opened 2016-04-07
php-cs-fixer issue


badams:
Running `php-cs-fixer fix . --level=psr2` generates a massive list of changes, and breaks unit tests (it changes the namespaces for tests from ` Yasumi\Tests\NewZealand` to  `Yasumi\tests\NewZealand`)

Are there some specific settings to use when running php-cs-fixer?

stelgenhof:
Oh really? I have never seen this on my own machine (using PHPStorm with php-cs-fixer integration). I just tried on a separate environment and noticed the same. Let me investigate what the difference is.

badams:
The difference isn't too bad, if you just run `php-cs-fixer fix src/ --level=psr2` it's mainly spacing/new line fixes along with a few function definitions that don't specify visibility.

if you run it on the tests directory also (as I did originally) it will update the namespaces on all the tests.. I think all you need to fix that issue is update following in composer.json (I've not tested this)

```
"autoload-dev": {
    "psr-4": {"Yasumi\\Tests\\": "tests/"}
  }
```

stelgenhof:
Yeah, I noticed that mostly it is comments and formatting. Oops! Not following PSR-2 as I stated :)
Let me check the "tests" change in composer.json

badams:
Yes, and oddly enough it doesn't do the same for other namespaces eg. `Yasumi\Provider\NewZealand`

stelgenhof:
I changed composer to 

```
"autoload-dev": {
    "psr-4": {"Yasumi\\tests\\": "tests/"}
  }
```

and HolidaysFilterTest.php also, and now phpunit runs without error.

Hmm... still wondering why this directory name was changed.

badams:
I guess if it doesn't hurt BC its not a bad thing to be more compliant - I trust php-cs-fixer knows what it's doing :)

stelgenhof:
Actually I did some tests, and I get the feeling it applies fixes that are not only PSR2 :(. For example it removes the '@package....' line from the comments and I checked that is part of the 'Symfony' level.

Looks like the option --level=psr2 doesn't mean exclusively PSR2.

For good measure I checked another package (Dingo API) and their composer.json contents for the unit tests is the same as mine.

badams:
Ah, I see.. that makes sense.

[...]

PR #16, opened 2016-04-08
15 php cs fixer issues

badams:
#15 Adding php-cs-fixer config, and changes from running said config


.php_cs
```diff
@@ -0,0 +1,80 @@
+<?php
+
+$finder = Symfony\CS\Finder\DefaultFinder::create()
+    ->in(__DIR__);
+
+$fixers = array(
```
stelgenhof: Sorry I haven't checked myself yet, but does this custom list cover PSR2? I just would like Yasumi to be PSR2 compliant. Any additional ones are fine to me, but not mandatory.

stelgenhof: Reviewed the PR and looking good! Mostly it is formatting/style corrections. However seems these are not necessarily PSR2 corrections, but nonetheless good. Let me merge it this weekend, as I want to check a few little things more

badams: I guess its a starting point, having the config in place means this can eventually be formalized, and the `php-cs-fixer` checks can be added to travis once that is done. 

There are a lot of formatting changes in this PR - I didn't even manage to review them all.

I've just been looking over the PSRs, and Fixers defined for them and this config still includes a lot of symfony cs fixers. I might revert the previous commit and re-run php-cs-fixer with only the PSR based fixers.

badams:
@stelgenhof I think I'm happy with this now :)

To summarize the changes in this PR

- I adjusted the list to only include PSR based fixers
- I've also added a shortcut command to composer `composer php-cs-fixer` which runs the fixers.
- I've php-cs-fixer now also runs on the CI server, meaning PRs will need to be PSR2 compliant before they can be merged.

badams:
Also, a note about the tests directory and namespaces, it appears to be part of PSR4

> The contiguous sub-namespace names after the "namespace prefix" correspond to a subdirectory within a "base directory", in which the namespace separators represent directory separators. The subdirectory name MUST match the case of the sub-namespace names.

Which is a requirement of PSR2 as far as I understand it

stelgenhof:
I think the contiguous sub-namespace requirements was the culprit I believe. Let me check once more and merge your PR

stelgenhof:
Would you mind changing the "tests" folder back to uppercase? I prefer the sub-namespaces to be consistent. If not, I can do it after merging the PR :)

As for `- ./vendor/bin/php-cs-fixer --diff --dry-run -v fix`, you said it will run on the CI server however this is just a dry-run. Any reason for that?

badams:
@stelgenhof would you want to change the actual directory from `tests` to `Tests`?  at least that way it does align still align with PSR4

stelgenhof:
Ah indeed. Then I would leave it as you did. No need to change. I will merge it

badams:
Alright, go ahead. I'll need to update my other pull request to meet the new CS requirements :)

PR #18, opened 2016-04-11
polish holidays & translation 

mpdx:


stelgenhof:
Thank you @mpdx ! Let me review the PR. Nice to have Poland added as well :)

stelgenhof:
There is a basic test file missing for the country/region itself. It should be called PolandTest.php and needs to contain tests which holidays you are expecting from the provider class. Check the other countries for an example.

mpdx:
@stelgenhof aye, mate! thanks for the info!

stelgenhof:
It seems there is no unit test for Corpus Christi. Can you add that one too?

PR #19, opened 2016-04-12
Added support for Julian calendar

zzenmate:
Adding support for Christian Holidays For Julian Calendar.


stelgenhof:
Thanks @black-bullet ! I will have a look and she when to merge it. Appreciate the PR for Julian Calender

stelgenhof:
I think this is really good. For example Greece is using Julian calendar for some of its holidays.

stelgenhof:
@black-bullet Looks good, although this only supports the years 1900 - 2099. It would also be good if this trait extends to the other eras as well. I prefer to have as good as possible implementation.

Appreciate your support!

PR #20, opened 2016-04-12
Fixed and secure Yasumi exposed function

R2c:
Fixed Yasumi::getProviders
Fixed & secure Yasumi::create

### DIFF
--- a/src/Yasumi/Yasumi.php
+++ b/src/Yasumi/Yasumi.php
@@ -12,7 +12,6 @@
 
 namespace Yasumi;
 
-use DirectoryIterator;
 use InvalidArgumentException;
 use RecursiveIteratorIterator;
 use RecursiveDirectoryIterator;
@@ -108,6 +107,16 @@ class Yasumi
     public static function getAvailableLocales()
     {
         return require __DIR__.'/data/locales.php';
+    }
+
+    /**
+     * @param $provider
+     *
+     * @return bool
+     */
+    public static function isProvider($provider)
+    {
+        return in_array($provider, self::getProviders());
     }
 
     /**
--- a/src/Yasumi/Yasumi.php
+++ b/src/Yasumi/Yasumi.php
@@ -72,7 +72,7 @@ class Yasumi
     {
         // Find and return holiday provider instance
         $providerClass = sprintf('Yasumi\Provider\%s', str_replace('/', '\\', $class));
-        if (!class_exists($providerClass)) {
+        if (!class_exists($providerClass) || $class === 'AbstractProvider') {
             throw new InvalidArgumentException(sprintf('Unable to find holiday provider "%s".', $class));
         }
 
--- a/src/Yasumi/Yasumi.php
+++ b/src/Yasumi/Yasumi.php
@@ -14,6 +14,8 @@ namespace Yasumi;
 
 use DirectoryIterator;
 use InvalidArgumentException;
+use RecursiveIteratorIterator;
+use RecursiveDirectoryIterator;
 use RuntimeException;
 use Yasumi\Exception\UnknownLocaleException;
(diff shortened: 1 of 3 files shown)


stelgenhof:
Super! Thanks the fixes. Before I merge your PR, can you explain a little bit more what the issue is with commit 5b4bd9d. I can't yet say what the issue here is :)

R2c:
For the commit  5b4bd9d. Before if you call getProviders you have : 
Belgium
Spain
France
... 
but it's not all providers
Now if you call getProviders you have : 
Belgium, 
France, 
France/HautRhin, 
...
Spain, 
Spain/Asturias
...

For the commit 5bba6a6, i think to use getProvider and check if the provider is valid. But for performance (check every provider if you need just one) it's not perfect. So i just disabled AbstractProvider. Any preference for this ? 

Before merge, i can add : isProvider method using getProviders

R2c:
Adding new commit with isProdiver.

src/Yasumi/Yasumi.php
```diff
@@ -111,6 +111,18 @@ public static function getAvailableLocales()
     }
 
     /**
+     * @param $provider
+     *
+     * @return bool
+     */
+    public static function isProvider($provider)
+    {
+        $providers = self::getProviders();
```
stelgenhof: Perhaps make this a one-liner? :)

stelgenhof:
Understood! I see now that the function getProviders is not working properly as it doesn't retrieve the subregions. Perfect I will merge your PR. Thanks for spotting this one!

stelgenhof:
Just had a final check and have one more question: what is the purpose of the isProvider function? It doesn't clearly say what type it is expecting. I assume it's a provider object, but in that case the create function will throw an "InvalidArgumentException" in the situation the class is not an existing holiday. Would that not be enough?

R2c:
This function works with your "getProviders" who return a array of provider names.
"isProvider" check if this string is a valid provider.

If i do that, it's ok ? 
```
    /**
     * Check if a provider name exists (France, Spain...)
     *
     * @param string $provider
     *
     * @return bool
     */
    public static function isProvider($provider)
    {
        return in_array($provider, self::getProviders());
    }
```

Otherwise, tell me what you want exactly and I will make it

stelgenhof:
Ah ok. My point was more that the function `isProvider` is public and was wondering if it would be used like that separately. It indeed makes sense for the getProviders function. If it is not supposed to be used exposed, perhaps better to change the visibility (protected/private).

R2c:
it's supposed to be used exposed.

```
if Yasumi::isProvider(test) {
   Yasumi::create...
} else {
   ....
}
```

Better and easier than 
```
Try {
   Yasumi::create...
catch (\InvalidArgumentException $e) {
   ...
}
```
No ? if you think is useless, i remove this function.

stelgenhof:
I see what you want to do with this function now. Could you perhaps tell what situation you would like to use this function? I am not sure I see the benefit of this function. The getProviders() function was created to get the entire list of holiday providers so you don't need an additional check like your suggestion :)

R2c:
In my case, i can add some holidays for just one country, not for all in one shot. So it's just a shortcut to avoid : 
```
if (in_array('France', Yasumi::getProviders()) {
   //do the job !
}
```

stelgenhof:
Ok :) Check this gist: https://gist.github.com/stelgenhof/037f16f97db1991a787f281610a1c3f0

R2c:
So i remove the last commit ?

stelgenhof:
At the moment I don't see a lot of benefit adding your suggested function. So if you could close the PR that would be nice. If you really would like to use a function like `isProvider` I suggest you extend the Yasumi class and add your own function :) Thanks again for the fixes and support. Appreciate it a lot.

PR #25, opened 2016-05-03
Welsh language translations (spoken in Wales, UK)

meigwilym:
I've translated the main holidays, and also added a file for St David's day which is Wales' patron Saint. The UK holiday provider ignore's saints' days, but I've added it for completeness.


stelgenhof:
Great! Appreciate it a lot. I will merge it asap. In any case, since Wales might have different holidays, I prefer to create a separate holiday class. This can be extended from the UK one so only the difference have to be implemented (See the examples of Spain and France).

PR #29, opened 2016-05-04
Update Yasumi::create to allow for the instantiation of external holiday providers

navarr:
I thought it was kind of silly that Yasumi can only load it's own holiday providers (using Yasumi::create), so I wrote this update that allows it to instantiate external ones.

External Holiday providers must be an instance of ProviderInterface.


stelgenhof:
Nice one! Could you perhaps provide/write an example how to use it? We can then use it in the documentation somewhere... :)

PR #30, opened 2016-05-04
Update Yasumi::getProviders for platform agnosticism

navarr:
Discovered while running tests on #29, Yasumi::getProviders does not work as expected on Windows!

This commit updates Yasumi::getProviders so that instead of hardcoding the unix-style / it instead listens for whatever PHP specifies as the DIRECTORY_SEPARATOR.

### DIFF
--- a/src/Yasumi/Yasumi.php
+++ b/src/Yasumi/Yasumi.php
@@ -123,9 +123,11 @@ class Yasumi
             return $providers;
         }
 
+        $ds = DIRECTORY_SEPARATOR;
+
         $extension     = 'php';
         $providers     = [];
-        $filesIterator = new \RecursiveIteratorIterator(new RecursiveDirectoryIterator(__DIR__ . '/Provider/'),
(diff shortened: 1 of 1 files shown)


stelgenhof:
Super! Was thinking about the same, but didn't have the time yet to test on Windows. My understanding is that DIRECTORY_SEPARATOR is not necessarily needed because PHP always converts `/` to the appropriate character in its file functions.

navarr:
This is what getProviders was returning on Windows:

stelgenhof:
Hmm. I see...

stelgenhof:
I tried on Windows and see also the 'Unix' style separators but didn't have any issues. Do you encounter any problems because of this?

navarr:
No problems in usage.  Only that the test failed because getProviders was returning full filepaths instead of provider names.

stelgenhof:
I didn't run the tests. Let me check again.

PR #31, opened 2016-05-04
Fix #28

### LINKED ISSUE #28: YasumiTest:textNext uses an out of bounds year
Running phpunit can result in an error in the test class `Yasumi\tests\Base\YasumiTest::testNext` in the rare situation that the year 10000 is applied. Yasumi only supports year numbers between 1000 and 9999.
PHPUnit Output: 

```
1) Yasumi\tests\Base\YasumiTest::testNext
InvalidArgumentException: Year needs to be between 1000 and 9999 (10000 given).
..../Users/sachatelgenhofoudekoehorst/src/azuya/Yasumi/tests/Base/YasumiTest.php:135
```

navarr:
By updating YasumiTest to call the factory with a max of 9998 so that when it is +1'ed it is within the limit.

### DIFF
--- a/tests/Base/YasumiTest.php
+++ b/tests/Base/YasumiTest.php
@@ -128,7 +128,7 @@ class YasumiTest extends PHPUnit_Framework_TestCase
     {
         $country = 'Japan';
         $name    = 'childrensDay';
-        $year    = Factory::create()->numberBetween(1949, 9999);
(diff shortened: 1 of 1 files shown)


stelgenhof:
Ah just beat you to it :) I push a changed to the master branch covering this. (Covers both the 'next' and the 'previous' methods).

navarr:
Shucks :p

PR #35, opened 2016-07-26
Added provider class with New York Stock Exchange holidays. Added test cases.

mgwebgroup:
Added class NYSE.php to the Provider directory. The class uses holidays observed by the New York Stock Exchange published on https://www.nyse.com/markets/hours-calendars

Added test classes 'tests/GoodFridayTest.php' and 'tests/NYSETest.php'

### DIFF
--- a/src/Yasumi/Provider/NYSE.php
+++ b/src/Yasumi/Provider/NYSE.php
@@ -1,0 +1,194 @@
+<?php
+/**
+ *  This file is part of the Yasumi package.
+ *
+ *  Copyright (c) 2015 - 2016 AzuyaLabs
+ *  Copyright (c) 2016 MGWebGroup
+ *
+ *  For the full copyright and license information, please view the LICENSE
+ *  file that was distributed with this source code.
+ *
+ * @author Alex Kay <alex110504@gmail.com>
+ */
+
+namespace Yasumi\Provider;
+
+use DateInterval;
+use DateTime;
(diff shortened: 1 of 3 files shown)


stelgenhof:
Thanks for the PR! Let me have a look at it this coming weekend.

stelgenhof:
Hi @mgwebgroup ,

I had a look at your PR and am not sure if this would be the right approach. NYSE wouldn't be a 'real' country/subregion since your class merely is a subset of the USA Holiday provider. I think a better approach would be to create it as a filter.

Cheers! Sacha

mgwebgroup:
Allright, I'll see what I can do. Do you have a preference for a filter name?

stelgenhof:
The filter name can be anything, as long as it is clear :) It would be very easy to implement: just check the dates that are considered part of the NYSE calendar in the 'accept' method of the filter.

The only small catch is that this filter will only work with the USA Holiday provider.

mgwebgroup:
I'll see what I can do. Thanks for the suggestion. I will be able to look into this after September 1st.

stelgenhof:
Sure no problem!

stelgenhof:
I have added a recipe how you can easily extend an existing holiday provider to create your own custom provider: https://azuyalabs.github.io/yasumi/cookbook/custom_provider

Note: you need to use the latest v1.8.0 release for this to work :)

PR #38, opened 2016-09-20
Adding Australia\Victoria region

brucealdridge:


brucealdridge:
Will add support for QLD/NSW/SA and maybe more later

stelgenhof:
Hi @brucealdridge, would you mind creating a PR for all Australia rather than bit by bit? That way I can put it in a release in it's entirety :)

splitice:
The national ones are contained in this PR, and thats exactly what I needed.

Thanks @brucealdridge

brucealdridge:
@stelgenhof national public holidays for Australia are in there, just not all states. 
Australia is weird - and some states have holidays on different dates (Queens Birthday, Labour Day) 

This PR at the moment contains the National holidays and all holidays for VIC. I'll get around to the other states but probably 2weeks away.

stelgenhof:
@brucealdridge Thanks for the update. I'll merge then merge this PR as soon as possible :)
Thanks for the contribution!

PR #39, opened 2016-09-20
Add working/business days to a date

brucealdridge:
Add the ability to calculate business days.
Ie 'Expect your parcel to arrive in 7 business days'
```php
$sevenBusinessDays = Yasumi::addWorkingDays('USA', 7, new \DateTime('2016-07-01', new \DateTimeZone('America/New_York')))
$sevenBusinessDays->format('Y-m-d');
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,13 @@ foreach ($official as $day) {
 // 'Whitmonday'
 // 'Christmas'
 // 'Boxing Day'
+
+// Calculate number of business/working days until a date
+$nextBusinessDay = Yasumi::addWorkingDays('USA', 1, new \DateTime('2016-07-01', new \DateTimeZone('America/New_York')))
+$nextBusinessDay->format('Y-m-d');
+// 2015-07-05
 ```
+
 
 Testing<a name="testing"></a>
 -------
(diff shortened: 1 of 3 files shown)


stelgenhof:
@brucealdridge Thanks for this one! I like it :) If I understand correctly this feature returns the first coming business in the next ## days, right? In that case, I prefer the function name to represent what it returns rather than the action ('adding days').

For completeness sake, we should also then have the possibility to look at past business days. Can't think of a use case right now, however would be good to have too.

brucealdridge:
An example would be customer service.
Instead of 
```php
echo "We will get back to you within 2 business days";
```
```php
echo "We will get back to you by ".
  Yasumi::addWorkingDays(
      'USA', 
      2, 
      new \DateTime('2016-07-01', new \DateTimeZone('America/New_York'))
   )->format('Y-m-d');
```

Fairly trivial to add `removeBusinessDays`, `nextWorkingDay` and `previousWorkingDay` but I see less of a use case for those

stelgenhof:
Yes, that makes sense. I was talking more semantically as for the function names. I think 'nextWorkingDay' and 'previousWorkingDay' make more sense. The functions then return either the next or previous working day based on the number of days given (as you coded). The default could be 1 day.

brucealdridge:
Sounds good, I'll adjust the naming

PR #40, opened 2016-09-25
Adds Slovakia

dakujem:
Adds holiday support for Slovakia
- holiday provider
- translations
- tests

Holidays in Slovakia are divided into bank holidays and national days.

### DIFF
--- a/tests/Slovakia/EasterMondayTest.php
+++ b/tests/Slovakia/EasterMondayTest.php
@@ -14,11 +14,10 @@
 
 namespace Yasumi\tests\Slovakia;
 
-use DateTime,
-	DateTimeZone,
-	Yasumi\Holiday,
-	Yasumi\tests\YasumiTestCaseInterface;
-
+use DateTime;
+use DateTimeZone;
+use Yasumi\Holiday;
+use Yasumi\tests\YasumiTestCaseInterface;
 
 /**
  * Class for testing a holiday in Slovakia.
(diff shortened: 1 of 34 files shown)


stelgenhof:
Thanks for the contribution! Let me have a look at the PR in the next few days :)

dakujem:
yw. no idea why the checks are failing. The unit tests passed on 5.6.

stelgenhof:
The unit test themselves are not failing. There is an additional check for PSR-2 coding standards which is failing. Please run `composer php-cd-fixer` before submitting your PR :)

tests/Slovakia/GoodFridayTest.php
```diff
@@ -0,0 +1,90 @@
...
+		$this->assertHoliday(self::REGION, self::HOLIDAY, $year, $expected);
+	}
+
+
+	/**
+	 * Returns a list of test dates used for assertion of the holiday defined in this test
+	 *
+	 * @return array list of test dates for the holiday defined in this test
+	 */
+	public function HolidayDataProvider()
```
stelgenhof: Do you think you can use a date generator rather than a fixed list of test dates? That would be much more sustainable.

tests/Slovakia/EasterMondayTest.php
```diff
@@ -0,0 +1,90 @@
...
+		$this->assertHoliday(self::REGION, self::HOLIDAY, $year, $expected);
+	}
+
+
+	/**
+	 * Returns a list of test dates used for assertion of the holiday defined in this test
+	 *
+	 * @return array list of test dates for the holiday defined in this test
+	 */
+	public function HolidayDataProvider()
```
stelgenhof: Do you think you can use a date generator rather than a fixed list of test dates? That would be much more sustainable.

stelgenhof:
For Slovakia, are there any holidays that have been established from a certain date or year? If so, it is good to add a condition to the calculation when such a holiday became in effect. Likewise if a holiday has been abolished. This allows for other users to still use past/new holidays for their use case.

Apart from the coding style, it all looks good!

dakujem:
From my knowledge, since 1993 when Slovakia became a souvereign country there have been no changes. Before we were a part of Czecho-Slovakia, I don't know how this should be handled. There were quite a few changes in the first part of the 20-th century.

dakujem:
Yay, I ran the cs fixer, but the build is still failing. I'm working on Windows, maybe that is a problem?

tests/Slovakia/GoodFridayTest.php
dakujem: I was inspired by other providers, they only use one date. I included 8 dates from past and future...

tests/Slovakia/EasterMondayTest.php
```diff
@@ -0,0 +1,88 @@
...
+        $this->assertHoliday(self::REGION, self::HOLIDAY, $year, $expected);
+    }
+
+
+    /**
+     * Returns a list of test dates used for assertion of the holiday defined in this test
+     *
+     * @return array list of test dates for the holiday defined in this test
+     */
+    public function HolidayDataProvider()
```
stelgenhof: Could you maybe use random dates (i.e. generator) rather than a fixed set of dates? That would be more sustainable and cover more dates.

stelgenhof:
If technically speaking the holidays didn't exist before the time Slovakia and Czech Republic became sovereign countries, I would then implement that the Slovakian holidays only are generated after 1993. This would account for Czech Republic then as well. (Before 1993 we should have a separate holiday provider for Czechoslovakia).

As for your errors, I don't use Windows so can't really help there. All is mostly built on non Windows platforms. At the end I can merge your PR and fix it on my side for you.

Cheers! Sacha

tests/Slovakia/GoodFridayTest.php
stelgenhof: I see. No problem. Usually it's better to have automatic test in stead of relying on manual fixtures :)

dakujem:
Well, then it would become quite complex - there would have to be at least 5 more providers for 20-th century only... As Slovakia was part of the Autro-Hungarian monarchy, then Czechoslovakia, then independent and separated, then Czechoslovakia again, then another Czechoslovakia, yet another, until finally independent since 1993. Not to mention changes in borders. I think that it's not the aim of the library to reflect history.

I suggest I check for the holidays celebrated since say end of WWII in the area of current Slovakia and incorporate them into the provider, provided I can find relevant information.

What is your opinion?

stelgenhof:
I agree that for the most part, future holidays are used. However there are some use cases where past holidays are required in certain calculations for example. In the Netherlands for example the national Queensday was changed date a few times and recently renamed to "Kingsday".

Although not indeed extremely important I would suggest for those holidays you know it only started from a specific year, implement the starting year. Otherwise leave it as is :)

dakujem:
Actually, I was unable to find any reliable information source of holidays before 1993. Some holidays were celebrated on the particular day with changed names, but from what I found, before 1989 there were many and much different holidays, since it was a different regime... So... it probably would require implementing separate provider anyway.

stelgenhof:
@dakujem Ok. Understood, makes sense. I'll merge your PR. It is always possible to address these situations at a later stage. Thanks again for the contribution!

PR #44, opened 2016-11-19
Added Portugal holiday provider

rvelhote:
Hey Sacha and maintainers,

I have created a provider for holidays in Portugal. The 2016 calendar in Portugal has 13 national holidays and I included only those 13. I plan on including some regional holidays as well in the future as well as past holidays that no longer exist (e.g. the 10th of June holiday had a different name between 1933 and 1973).

A few of the holidays present in 2016 were abolished in 2014 and restored this year (2016) and that factor is taken into account in the provider.

These three holidays are:
- Corpus Christi
- Restoration of Independence (1st of December)
- All Saints Day

I await your feedback.

Thank you and best regards,
Ricardo Velhote

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ Yasumi currently supports 59 countries (including sub-regions):
 * New Zealand
 * Norway
 * Poland
+* Portugal
 * Slovakia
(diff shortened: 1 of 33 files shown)


stelgenhof:
Thanks for this PR! Let me have a look and merge it if all is fine :)

PR #53, opened 2017-01-08
Update secondChristmasDay.php

AngelinCalu:
Boxing Day concept does not exist in the US. This should be instead "Second Christmas Day".


stelgenhof:
After having a closer look, the USA doesn't observe a second Christmas Day. The translation in this file for the 'en_US' does not relate to a (possible) second Christmas Day: it is merely the 'translation' of it for other countries where a second Christmas Day is observed. Nonetheless, if it makes sense for the en_US locale to have it changed, then we can adjust this :)

AngelinCalu:
Yes. In my case I am only using the english version of the holidays for several countries from Europe.  Probably most of the users are interested in the english version as well.
As I see it, all holidays should have a common (english) translation.

PR #54, opened 2017-02-04
Require php calendar extension to prevent breaking sites

Jarlskov:
I've accidentally taken down a production site because the server did not have the calendar extension installed, and I was not aware that I needed it (or that it existed actually).
Since it's a hard dependency for the library to function, that should be more obvious :-)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
     }
   ],
   "require": {
-    "php": ">=5.6.0"
+    "php": ">=5.6.0",
+    "ext-calendar": "*"
   },
   "require-dev": {
     "phpunit/phpunit": "~4.8|~5.7",


stelgenhof:
@Jarlskov Sorry to hear that your site was taken down because of this :( You are correct that it is not clear that the calendar extension is actually required. If I recall, I wanted to stay away from such dependency but think I ended up using it anyway.

Your PR is good in that it will enforce the use of the calendar extension. Now that I am thinking of it, I think I want to provide an implementation of the `easter_days` function within Yasumi that would replace the need for the extension after all.

Jarlskov:
No worries, that's our own fault. We were able to roll back quickly, and it showed some flaws in our test procedures :-)

I'm not sure why you would want to maintain existing functionality, but that's up to you how you want to solve it, but I think it would make sense to add something like this until you get around to implementing the easter_days functions. I don't believe this PR should cause any BC breaks.

stelgenhof:
Ah yes, I see. Merging this PR will indeed not cause any BC issues. Implementing a easter_days function as part of Yasumi isn't that difficult and planning on releasing it pretty soon :)

stelgenhof:
Fallback easter_days calculation implemented in case the PHP calendar extension is not loaded. (See PR #55)

PR #56, opened 2017-02-23
Adding Swiss provider

qligier:
This PR adds common holidays of Switzerland (including the sub-regions Aargau, Appenzell Ausserrhoden, Appenzell Innerrhoden, Basel Landschaft, Basel Stadt, Bern, Fribourg, Geneva, Glarus, Grisons, Jura, Lucerne, Neuchatel, Nidwalden, Obwalden, Schaffhausen, Schwyz, Solothurn, St. Gallen, Thurgau, Ticino, Uri, Valais, Vaud, Zug and Zurich).

Except for the Swiss national day, holidays are to be chosen by cantons, not by the government so a provider for each canton is provided.

The PR includes the providers, unit tests and translation of common holidays.


stelgenhof:
@qligier Thank you so much for this! Must have been quite a job. Let me have a look and merge your PR if all looks fine 👍

qligier:
It has been quite hard because of the fragmentation of holidays [visible in this doc](https://www.tcs.ch/fr/assets/voyages-camping/infos-touristiques/themes/CHE_jours_feries_17.pdf) (in blue, holidays that only exist in sub-sub-regions).

I've implemented all holidays except the Swiss national day as other. They could also be set as national or we could create a regional type.

stelgenhof:
After briefly checking your PR, I realize that the type 'NATIONAL' is not quite appropriate. My intention was this one to reflect the official holidays irrespective of the region/subregion. I think I am going to change the naming for that so it better reflects this situation.

qligier:
Renaming _national_ to _official_ could be a good thing (even if _bank_ and _seasonal_ holidays also are official)? Differentiate between national and regional holidays is maybe not that important. That could also be useful for Australia, France, Germany and Spain.

stelgenhof:
Yes, the main purpose of the different types was to differentiate between official/government approved holidays and other (observed, seasonal, others, etc.) type of holidays. Perhaps that needs a refactoring :)

PR #60, opened 2017-04-18
Added missing translation (de_DE) for Easter Sunday + Whitsunday

IceShack:


stelgenhof:
Super! Thank you for adding these.

Issue #61, opened 2017-05-31
add yasumi creation by ISO3166-2 code


huehnerhose:
I want to get holidays based on dynamically gathered region informations. The ISO3166-2 codes I can get out of various map-api. 
As far as I can see the `const ID` inside the providers corresponce to ISO3166-2 and is used as index for the `getProviders()` return array. 

Therefore I made a simple [wrapper](https://github.com/InSitu-Software/yasumi/commit/472aba9c7c958ab1bccd0083a5fba4133be1e813). There aren't any test for this wrapper implemented and I only "tested" it against my usage. (Germany and Subdivisions of Germany).

So this is kind of a "Feature Request" / "Pull Request" mix :)

stelgenhof:
Very nice! Some others have asked a similar feature as currently you need to know the exact name of the Holiday provider. Would you be able to create a PR?

stelgenhof:
Closing issue. Resolved by PR #62

PR #62, opened 2017-06-06
add yasumi creation by ISO3166-2 code

huehnerhose:
See [issue](https://github.com/azuyalabs/yasumi/issues/61)

I needed the option to create a yasumi instance by ISO3166-2 Code. As far as I can see, the `const ID` in the providers corresponces to this code. Therefore I created this simple wrapper.

I wasn't able to create test etc. for this wrapper. I think this should be added before mergen

### DIFF
--- a/src/Yasumi/Yasumi.php
+++ b/src/Yasumi/Yasumi.php
@@ -189,8 +189,8 @@ class Yasumi
      * Create a new holiday provider instance.
      *
      * A new holiday provider instance can be created using this function. You can use one of the providers included
-     * already with Yasumi, or your own provider by giving the name of your class in the first parameter. Your provider
-     * class needs to implement the 'ProviderInterface' class.
+     * already with Yasumi, or your own provider by giving the 'const ID', corresponding to the ISO3166-2 Code, set in
+     * your class in the first parameter. Your provider class needs to implement the 'ProviderInterface' class.
      *
      * @param string $iso3166_2  ISO3166-2 Coded region, holiday provider will be searched for
      * @param int    $year   year for which the country provider needs to be created. Year needs to be a valid integer
@@ -200,7 +200,7 @@ class Yasumi
      * @throws RuntimeException         If no such holiday provider is found
      * @throws InvalidArgumentException if the year parameter is not between 1000 and 9999
      * @throws UnknownLocaleException   if the locale parameter is invalid
-     * @throws InvalidArgumentException if the holiday provider for the given country does not exist
+     * @throws InvalidArgumentException if the holiday provider for the given ISO3166-2 code does not exist
      *
      * @return AbstractProvider An instance of class $class is created and returned
      */
--- a/src/Yasumi/Yasumi.php
+++ b/src/Yasumi/Yasumi.php
@@ -186,6 +186,43 @@ class Yasumi
     }
 
     /**
+     * Create a new holiday provider instance.
+     *
+     * A new holiday provider instance can be created using this function. You can use one of the providers included
+     * already with Yasumi, or your own provider by giving the name of your class in the first parameter. Your provider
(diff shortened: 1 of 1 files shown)


src/Yasumi/Yasumi.php
```diff
@@ -186,6 +186,44 @@ public static function create($class, $year = null, $locale = self::DEFAULT_LOCA
...
+     * @throws UnknownLocaleException   if the locale parameter is invalid
+     * @throws InvalidArgumentException if the holiday provider for the given country does not exist
+     *
+     * @return AbstractProvider An instance of class $class is created and returned
+     */
+    public static function createByISO3166_2($iso3166_2, $year = null, $locale = self::DEFAULT_LOCALE)
+    {
+        $providers = self::getProviders();
+
+        $class = isset($providers[$iso3166_2])
```
stelgenhof: You can simplify this by combining this and the '`if ($class === false)`' into one. Then no need for the intermediate variable `$class` and ternary check.

src/Yasumi/Yasumi.php
```diff
@@ -186,6 +186,44 @@ public static function create($class, $year = null, $locale = self::DEFAULT_LOCA
...
+     *
+     * @param string $iso3166_2  ISO3166-2 Coded region, holiday provider will be searched for
+     * @param int    $year   year for which the country provider needs to be created. Year needs to be a valid integer
+     *                       between 1000 and 9999.
+     * @param string $locale The locale to use. If empty we'll use the default locale (en_US)
+     *
+     * @throws RuntimeException         If no such holiday provider is found
+     * @throws InvalidArgumentException if the year parameter is not between 1000 and 9999
+     * @throws UnknownLocaleException   if the locale parameter is invalid
+     * @throws InvalidArgumentException if the holiday provider for the given country does not exist
```
stelgenhof: This needs to be corrected as the parameter is not 'country' but 'ISO3166-2'.

src/Yasumi/Yasumi.php
```diff
@@ -186,6 +186,44 @@ public static function create($class, $year = null, $locale = self::DEFAULT_LOCA
     }
 
     /**
+     * Create a new holiday provider instance.
+     *
+     * A new holiday provider instance can be created using this function. You can use one of the providers included
```
stelgenhof: This description needs to be edited since this function is using the 'ISO3166-2' code as the input parameter.

huehnerhose: ✔️

src/Yasumi/Yasumi.php
huehnerhose: ✔️

src/Yasumi/Yasumi.php
huehnerhose: This one I am not sure about. 
I can see that the assign of `$providers` is kind of unnecessary since the `getProviders` uses a static internal variable preventing the repetitive directory scanning. 

But I am not convinced that something like: 
    
        if ( false === isset( self::getProviders()[$iso3166_2] ) ) {
            throw new InvalidArgumentException(sprintf('Unable to find holiday provider by ISO3166-2 "%s".', $iso3166_2));
        }

        return self::create(
            self::getProviders()[$iso3166_2],
            $year,
            $locale
        );

benefits the readability of the code

stelgenhof: Yes, I see what you mean. What about this way:

```
$providers = self::getProviders();

if ( false === isset( $providers[$iso3166_2] ) ) {
        throw new InvalidArgumentException(sprintf('Unable to find holiday provider by ISO3166-2 "%s".', $iso3166_2));
    }

    return self::create(
        $providers[$iso3166_2],
        $year,
        $locale
    );
```

huehnerhose: deal :)

PR #68, opened 2017-10-11
Fix Saxony (Germany) test for other holidays

lukosius:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Generated random year (initial 1990) can sometimes be lower than
established holiday (Repentance and Prayer Day) date 1995 which causes
test to fail.

### DIFF
--- a/tests/Germany/Saxony/SaxonyTest.php
+++ b/tests/Germany/Saxony/SaxonyTest.php
@@ -82,7 +82,13 @@ class SaxonyTest extends SaxonyBaseTestCase
      */
     public function testOtherHolidays()
     {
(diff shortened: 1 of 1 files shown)


stelgenhof:
Thank you! I actually already discovered this but haven't had the chance to correct it. Some unit tests still might need to get adjusted in these kind of situations. Not all of them have been reviewed yet.

Thanks for the PR!

stelgenhof:
Just have resolved this in other commit. The code for Saxony had to be adjusted as Repentance is an official holiday. So included also your code change :). Thanks for the contribution!

PR #74, opened 2017-10-22
Add fr_BE locale

jief:


src/Yasumi/data/translations/goodFriday.php
```diff
@@ -26,8 +26,9 @@
     'en_ZA' => 'Good Friday',
     'es_ES' => 'Viernes Santo',
     'fi_FI' => 'Pitkäperjantai',
+    'fr_BE' => 'Vendredi Saint',
     'fr_CH' => 'Vendredi Saint',
-    'fr_FR' => 'Vendredi saint',
+    'fr_FR' => 'Vendredi Saint',
```
stelgenhof: This change cause the unit test for France to fail. Please correct the corresponding unit test for Good Friday (France) as well.

jief: Done

Issue #75, opened 2017-11-02
create $year should by typed


huehnerhose:
I ran into the problem, that I call `create` (or `createByISO3166_2` in my case) like this:

    $holidays = Yasumi\Yasumi::createByISO3166_2(
				$region,
				$date->format("Y"),
				"de_DE"
			);

Since `$date->format("Y") ` gives me a string the later comparisons inside Yasumi failed 

    if ($year === 2017)

I see three ways solvings this:
1) TypeHinting. My prefered solution
2) Casting / Validating inside the create function (that what type hinting is for)
3) changing the `===` to type unaware `==` (Not a solution at all :) )

I saw the documentation states integer as `$year` but I imagine this to be a stupid trap you can walk into unaware and php offers a nice solution for it.

stelgenhof:
Thank you for catching this. Looking at the three possibilities, I am also in favor of type hinting.

stelgenhof:
Yasumi v2 has been released. This has been resolved now.

PR #76, opened 2017-11-02
TypeHinting for public functions

huehnerhose:
Like mentioned in #75 I would prefer type hinting as solution. 
Since TypeHinting doesn't allow default parameter values other than null, we have to default them manually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 ## [Unreleased]
 
 ### Added
-- Added Holiday Provider for Lithuania. [\#67](https://github.com/azuyalabs/yasumi/pull/67) ([lukosius](https://github.com/lukosius))
+- Holiday Provider for Latvia. [\#70](https://github.com/azuyalabs/yasumi/pull/70) ([lukosius](https://github.com/lukosius))
+- Holiday Provider for Lithuania. [\#67](https://github.com/azuyalabs/yasumi/pull/67) ([lukosius](https://github.com/lukosius))
 - Sometimes it is more convenient to be able to create a Yasumi instance by ISO3166 code rather than Yasumi's Holiday Provider name. A new function `createByISO3166_2` has been added to allow for that. [\#62](https://github.com/azuyalabs/yasumi/pull/62) ([huehnerhose](https://github.com/huehnerhose))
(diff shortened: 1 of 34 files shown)


huehnerhose:
If I am not mistaken, reverting this to 12645bc should pass on php5.6 as well.

src/Yasumi/Yasumi.php
```diff
@@ -120,7 +120,7 @@ public static function nextWorkingDay(string $class, DateTime $startDate, int $w
         $date = $startDate instanceof DateTime ? clone $startDate : new DateTime($startDate);
 
         // Default to 1
-        $workingDays = is_null($workingDays) ? 1 : $workingDays;
+        $workingDays = $workingDays ?? 1;
```
stelgenhof: The Null Coalescing Operator was introduced in PHP7. At the moment Yasumi still supports PHP5 so this will cause BC issues.

huehnerhose: Yes. At first I didn't use null coalescing, but than I thought it would look nicer and since all my projects depend on php7.1, I did't thought twice.

huehnerhose:
I reverted this and switched back to the ternary version. But Travis CI doesn't seem to like this very much. I ran the test locally with success. 

Honestly I am puzzled by the errors raised, especially on the 5.6 environment claiming to expect a `Yasum\string`.

stelgenhof:
Your type hint uses 'string' in the current namespace 'Yasumi', which doesn't exist of course. Also 'string' as a type was only introduced as of version 7 :( See: http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration

huehnerhose:
Urg... didn't realized that type hinting first came for not scalar types. Sorry. 

I don't know how important php5 support ist for you / your users. If you consider it valuable we can make the check "by hand" inside the functions and manually throwing exceptions.

huehnerhose:
I wanted to draft a php5 version 
https://github.com/InSitu-Software/yasumi/tree/feature/typehinted5
but now all tests fail with

    Yasumi\tests\Ukraine\SecondInternationalWorkersDayTest::testSecondInternationalWorkersDay with data set #9 ('1989', DateTime Object (...))
    InvalidArgumentException: integer expected, string given

😅

stelgenhof:
I'll create a branch for the PHP7 development. It's something I've been planning but haven't gotten around to. Maybe better to make all your changes there. In the near future I'd like to end PHP5 support and continue to use PHP7.

stelgenhof:
PHP7 branch created :). Please go ahead and use that for any PHP7 related enhancements.

stelgenhof:
I've updated the travis configuration so it should work now.

PR #88, opened 2018-01-05
Add PHPStan static analysis tool to travis

lukosius:
Code fixes: #87


stelgenhof:
Looks like the output of PHPStan causes the build to fail.

lukosius:
Yes. The #87 should be merged first which fixes the errors reported by PHPStan, then this PR will build successfully.

PR #100, opened 2018-03-07
easter sunday and pentecost sunday are not nationwide german holidays

TalonTR:
They are holidays only in Brandenburg, see https://de.wikipedia.org/wiki/Gesetzliche_Feiertage_in_Deutschland#%C3%9Cbersicht_aller_gesetzlichen_Feiertage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Fixed
 - Fixed spelling issue in the Swedish translation. [\#97](https://github.com/azuyalabs/yasumi/pull/97)
+- Fixed German Easter Sunday and Pentecost Sunday holidays (not nationwide, only in Brandenburg). [\#100](https://github.com/azuyalabs/yasumi/pull/100)
 
 ### Removed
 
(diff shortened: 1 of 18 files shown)


stelgenhof:
Thanks for the PR!

The source for defining the German Public Holidays was mainly Wikipedia: https://en.wikipedia.org/wiki/Public_holidays_in_Germany. This page list Easter Sunday and Pentecost as celebrated in all states. Comparing it with your source, this one is then incorrect?

Cheers! Sacha

crossan007:
I just discovered that Easter does not show up in the 'USA' provider as well.

stelgenhof:
@crossan007 It's correct it is not included as currently the USA provider focuses on Federal Holidays only. Of course, non-federal/other holidays can be added as well :)

Cheers! Sacha

TalonTR:
The table from your referenced english Wikipedia page doesn't list pentecost sunday and easter sunday for any german state. Below the table it states:

> In addition the state of Brandenburg has formally declared Easter Sunday and Pentecost Sunday as public holidays. As these are Sundays anyway, they have been left away by the other states, nor counted in the table above (the state of Hesse even declared all Sundays public holidays)."

This is exactly the same as within the german Wikipedia.

stelgenhof:
@TalonTR Perfect! I must have overlooked that.

TalonTR:
Didn't i already do that? All German Easter Sunday and Pentecost Sunday Tests are modified.

TalonTR:
Done :)

PR #101, opened 2018-03-07
Ignore time part and timezone

### LINKED ISSUE #78: Between filter wants me to work at new year
# Problem
    $filtered = $yasumi->between(
        new DateTime(($year).'-01-01'),
        new DateTime(($year + 1).'-01-01')
    );

Returns the 2017-01-01 as holiday only if the `DateTime` Timezone is the same as the local timezone, not if I use an DateTime with UTC.

# Background
Internally I use a wrapped DateTime class fixed to UTC with Hours:minutes to 0 to identify a single day. My local timezone is "Europe/Berlin". So DateTime defaults to `2017-01-01 00:00:00+1` and my class defaults to `2017-01-01 00:00:00Z`. 

So these two describe the following "moments in spacetime written without any offset"
DateTime (Europe/Berlin): `2016-12-31 23:00:00Z`
MyClass: `2017-01-01Z`

Yasumi creates the holidays based on the timezone given in the provider class. So germanys new year is: `2017-01-01 00:00:00+1` or `2016-12-31 23:00:00Z`. 
The between filter function now checks if 

    $holiday >= $this->start_date && $holiday <= $this->end_date

which must be denied, when `$this->start_date = MyUTCDay('2017-01-01')` because

    2016-12-31 23:00:00Z >= 2017-01-01 00:00:00Z

is false

# Discussion
I can understand the way you think, but I would like to advocate for the following idea:

If the holiday intersects with the given between barrier include it into the result set. 

On the pro side I can be sure to get _all_ holidays in my time frame. So I could ask the result set for any moment inside its timeframe if this moment is a holiday, regardless of day. 

On the against side I could end in getting too many holidays which results in unesecarry unwanted holidays in the result set. 

# Conclusion
I'd like to advocate for an option to get an inclusive between filter _or_ an option to query for a day without a timezone which then defaults to the provider class timezone.

c960657:
Fixes #78.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Fixed
 - Fixed spelling issue in the Swedish translation. [\#97](https://github.com/azuyalabs/yasumi/pull/97)
+- Fixed BetweenFilter to ignore time part and timezone. [\#101](https://github.com/azuyalabs/yasumi/pull/101)
 
 ### Removed
 
--- a/src/Yasumi/Filters/BetweenFilter.php
+++ b/src/Yasumi/Filters/BetweenFilter.php
@@ -30,12 +30,12 @@ use Iterator;
 class BetweenFilter extends FilterIterator implements Countable
 {
     /**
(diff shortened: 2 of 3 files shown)


c960657:
Sorry, I forgot that. Fixed :-)

PR #103, opened 2018-03-17
Get summer/winter time from PHP timezone library

c960657:
The date of transition to/from daylight saving time, and whether daylight saving time has been used at all, has changed over the years. This information is available in the tz (Olson) database that is included in PHP.

This PR makes Yasumi use the tz database for calculating the transitions. This works for any timezone that observes daylight saving time.

It seems that PHP currently only supports years until 2037 (when Unix timestamps exceed 31 bits), but this will likely be fixed in due time, if daylight saving time exists at all at that time.

### DIFF
--- a/src/Yasumi/Provider/CommonHolidays.php
+++ b/src/Yasumi/Provider/CommonHolidays.php
@@ -355,4 +355,98 @@ trait CommonHolidays
             $type
         );
     }
+
+    /**
+     * Calculates daylight saving time transitions.
+     *
+     * Daylight saving time is the practice of advancing clocks by one hour during summer months so evening daylight lasts even longer, while sacrificing normal sunrise times.
+     *
+     * The date of transition between standard time and daylight saving time differs from country to country and sometimes from year to year. Most countries outside Europe and North America do not observe daylight saving time.
+     *
+     * On the northern hemisphere, summer time starts around March/April. On the southern hemisphere it happens 6 months later.
+     *
+     * @param int    $year     the year for which Easter needs to be calculated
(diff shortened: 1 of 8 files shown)


stelgenhof:
Thanks for the PR! I think it is a good one, since it can now be used by other Holiday Providers as well. Always love a good refactor :)

src/Yasumi/Provider/CommonHolidays.php
```diff
@@ -355,4 +355,98 @@ public function armisticeDay(
...
+    protected function calculateSummerWinterTime($year, $timezone, $summer)
+    {
+        $zone = new DateTimeZone($timezone);
+
+        $transitions = $zone->getTransitions(mktime(0, 0, 0, 1, 1, $year), mktime(23, 59, 59, 12, 31, $year));
+
+        $transition = array_shift($transitions);
+        $dst = $transition['isdst'];
+
+        foreach ($transitions as $transition) {
```
stelgenhof: It seems you rely on the fact that the `getTransitions` function returns the transitions sorted by time, correct? If returned randomly, I think getting the transitions for summer and wintertime then might not work this way.

c960657: AFAICT from the PHP source code, we can rely on the dates being returned in chronological order. It seems the underlying data structure is also sorted by date.

https://github.com/php/php-src/blob/ef255c9f0f2c2ad1ea224ed215edb00f5bc205bd/ext/date/php_date.c#L4013

c960657:
Travis is happy now :-)

PR #104, opened 2018-04-21
Add observed holidays for Denmark

c960657:
The provider for Denmark currently supports what are usually known as holidays (_hellidage_).

Other days days are also observed:
1) _Lukkeloven_ (the Opening Hours Act) is the law defines when shops are allowed to stay open. In addition to the above mentioned holidays, the law prohibits most shops from staying open on [Constitution Day](https://en.wikipedia.org/wiki/Constitution_Day_(Denmark)) (5 June), Christmas Eve, New Year's Eve (partially) after 15:00, so shop staff have these days off. Others usually do not work on Christmas Eve and New Year's Eve, though this is not covered by any legislation.
2) Labour Day (1 May) is not mentioned in any legislation, but for people working under a trade union agreement (80% of the work force does that), either Constitution Day or Labour Day (depending on local/industry agreements) is a day off (except of course for police officers, nurses, taxi drivers etc. who also work on weekends).

Also, Denmark observes daylight saving time.

### DIFF
--- a/src/Yasumi/Provider/Latvia.php
+++ b/src/Yasumi/Provider/Latvia.php
@@ -57,7 +57,7 @@ class Latvia extends AbstractProvider
         $this->addHoliday($this->christmasEve($this->year, $this->timezone, $this->locale, Holiday::TYPE_OFFICIAL));
         $this->addHoliday($this->christmasDay($this->year, $this->timezone, $this->locale));
         $this->addHoliday($this->secondChristmasDay($this->year, $this->timezone, $this->locale));
-        $this->addNewYearsEve();
+        $this->addHoliday($this->newYearsEve($this->year, $this->timezone, $this->locale));
     }
 
     /**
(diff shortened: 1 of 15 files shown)


src/Yasumi/Provider/Latvia.php
```diff
@@ -57,7 +57,7 @@ public function initialize()
         $this->addHoliday($this->christmasEve($this->year, $this->timezone, $this->locale, Holiday::TYPE_OFFICIAL));
         $this->addHoliday($this->christmasDay($this->year, $this->timezone, $this->locale));
         $this->addHoliday($this->secondChristmasDay($this->year, $this->timezone, $this->locale));
-        $this->addNewYearsEve();
+        $this->addHoliday($this->newYearsEve($this->year, $this->timezone, $this->locale));
```
stelgenhof: Shouldn't this change be excluded from this PR since it isn't related to Denmark?

c960657: Both Denmark and Latvia (and many other countries) celebrate New Year's Eve, so I moved `newYearsEve ()` to `CommonHolidays` for easier reuse. So there is no reason to keep a duplicate implementation in the Latvia class.

PR #105, opened 2018-04-22
assertNotHoliday() skips rest of test

c960657:
If you add `assertNotHoliday()` to a test function, the rest of the test is skipped, because it intentionally throws an exception. This is rather surprising/confusing.


stelgenhof:
Thanks for catching this. As a matter of fact, the `assertFalse` (line 133) shouldn't have been there. The objective of this function is to assert that a holiday doesn't exist. In such a case always 'null' is returned, therefore making this assertion irrelevant.

PR #112, opened 2018-06-21
Australian Public Holidays

Milamber33:
Proper handling for (just about) all public holidays in every Australian state and territory.

### DIFF
--- a/src/Yasumi/Provider/Australia.php
+++ b/src/Yasumi/Provider/Australia.php
@@ -8,6 +8,7 @@
  * file that was distributed with this source code.
  *
  * @author Sacha Telgenhof <stelgenhof@gmail.com>
+ * @author William Sanders <williamrsanders@hotmail.com>
  */
 
 namespace Yasumi\Provider;
@@ -35,7 +36,6 @@ class Australia extends AbstractProvider
     /**
      * Initialize holidays for Australia.
      *
-     * @throws \Yasumi\Exception\InvalidDateException
      * @throws \InvalidArgumentException
      * @throws \Yasumi\Exception\UnknownLocaleException
      * @throws \Exception
@@ -43,8 +43,8 @@ class Australia extends AbstractProvider
     public function initialize()
     {
         // Official Holidays
-        $this->calculateAustraliaDay();
         $this->calculateNewYearHolidays();
+        $this->calculateAustraliaDay();
         $this->calculateAnzacDay();
 
         // Add Christian holidays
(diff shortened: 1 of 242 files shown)


stelgenhof:
Thanks! Sorry for not replying sooner. There is no need to close PR's if something needed to change.
I'll review your PR and merge it if it looks good :)

tests/Australia/ACT/QueensBirthdayTest.php
```diff
@@ -52,13 +51,12 @@ public function testHoliday($year, $expected)
...
     }
 
-
     /**
      * Returns a list of test dates
      *
      * @return array list of test dates for the holiday defined in this test
      */
-    public function HolidayDataProvider(): array
+    public function HolidayDataProvider()
```
stelgenhof: Yasumi is supporting PHP7 going forward, so please use/keep any return type declarations.

src/Yasumi/Provider/Australia.php
```diff
@@ -76,73 +75,84 @@ public function calculateAustraliaDay()
...
+     * @param string		  $type
      *
      * @throws \Yasumi\Exception\InvalidDateException
      * @throws \InvalidArgumentException
      * @throws \Yasumi\Exception\UnknownLocaleException
      * @throws \Exception
      */
     public function calculateHoliday(
-        string $shortName,
+        $shortName,
```
stelgenhof: Yasumi is supporting PHP7 going forward, so please use/keep any return type declarations.

src/Yasumi/Provider/Australia.php
```diff
@@ -76,73 +75,84 @@ public function calculateAustraliaDay()
...
      * Function to simplify moving holidays to mondays if required
      *
-     * @param string    $shortName
-     * @param array     $names
-     * @param \DateTime $date
-     * @param bool      $moveFromSaturday
-     * @param bool      $moveFromSunday
+     * @param string          $shortName
+     * @param array           $names
+     * @param string|DateTime $date
```
stelgenhof: Please use the DateTime type only.

Milamber33: Okay, that... breaks a lot. It will take me a few hours to make all the other changes to stop using strings everywhere.

tests/Australia/ACT/QueensBirthdayTest.php
Milamber33: Sure, but this is probably going to be needed in a lot of the other ones I've written. I copied and pasted a lot. I'll go through and try to get them all.

Milamber33:
Okay, got those ones you pointed out, and all the other instances where they were copied. There could well be more missing type declarations, I was working based off the current release when adding the extra files, and I used copy+paste a lot. Just let me know if you see any and I'll deal with them. :)

stelgenhof:
Thanks a lot! 

My apologies if I haven't been clear. Indeed the current stable release supports PHP 5.6 and greater. The development branch only supports PHP7 and greater and will be the basis for the next v2.0 release. I haven't documented/communicated that properly (yet).

PR #113, opened 2018-08-08
Changing Japan's holiday for the 2020 Olympic Games.

cookie-maker:
It corresponds to the Japanese holiday of 2020.

In 2020
MarineDay: 2020/7/23
HealthAndSportsDay: 2020/7/24
MountainDay: 2020/8/10

http://www8.cao.go.jp/chosei/shukujitsu/gaiyou.html

### DIFF
--- a/tests/Japan/HealthAndSportsDayTest.php
+++ b/tests/Japan/HealthAndSportsDayTest.php
@@ -33,6 +33,20 @@ class HealthAndSportsDayTest extends JapanBaseTestCase implements YasumiTestCase
     const ESTABLISHMENT_YEAR = 1996;
 
     /**
+     * Tests Health And Sports Day in 2020. Health And Sports Day in 2020 is July 24th for the Olympic Games.
+     */
+    public function testHealthAndSportsDayIn2020()
+    {
+        $year = 2020;
+        $this->assertHoliday(
+            self::REGION,
+            self::HOLIDAY,
+            $year,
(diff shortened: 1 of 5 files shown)


tests/Japan/HealthAndSportsDayTest.php
```diff
@@ -32,6 +32,20 @@ class HealthAndSportsDayTest extends JapanBaseTestCase implements YasumiTestCase
      */
     const ESTABLISHMENT_YEAR = 1996;
 
+    /**
+     * Tests Health And Sports Day in 2020. Health And Sports Day in 2020 is July 24th for the Olympic Games.
+     */
+    public function testHealthAndSportsDayIn2020()
+    {
+        $year = 2020;
```
stelgenhof: You don't need the variable $year here as it is only used once :)

tests/Japan/MarineDayTest.php
```diff
@@ -32,6 +32,20 @@ class MarineDayTest extends JapanBaseTestCase implements YasumiTestCaseInterface
      */
     const ESTABLISHMENT_YEAR = 1996;
 
+    /**
+     * Tests Marine Day in 2020. Marine Day in 2020 is July 23th for the Olympic Games.
+     */
+    public function testMarineDayIn2020()
+    {
+        $year = 2020;
```
stelgenhof: You don't need the variable $year here as it is only used once :)

cookie-maker:
Thank you for confirmation.
Updated the ChangeLog and commit into the develop brach.

PR #122, opened 2019-01-13
Remove locale from Holiday and Provider

c960657:
I suggest removing the locale from `Holiday` and `Provider` instances. The locale is not used by these entities themselves except when generating a name.

Instead I suggest adding a configurable global locale, and the possibility to override this in calls to `Holiday::getName()`.

This approach is used e.g. by the [PHP intl extension](http://php.net/manual/en/locale.setdefault.php), the [Symfony Translator](https://api.symfony.com/2.7/Symfony/Component/Translation/Translator.html), [Drupal's t() function](https://api.drupal.org/api/drupal/core%21includes%21bootstrap.inc/function/t/8.4.x) and the [Punic](https://punic.github.io/#Data-class) library.

This will make a clearer separation of concerns as well as reduce the amount of boilerplate code in all the providers.

The change is not backwards compatible, so this is 3.x material.

I have not implemented this for all providers yet. I'd appreciate some feedback on this PR before I proceed with fixing the remaining 9904 failing tests :-)

My reason for suggesting this is to pave the way for another PR that will allow custom locale fallbacks instead of the hardcoded fallback to English. E.g. if you are displaying a Swedish calendar with a Danish locale, not all Swedish holidays may have a Danish translation, but because the two languages are so closely related, you would rather want to fall back to Swedish than English. This change will be much easier if both locale and fallback locales wont have to be passed through all the providers.

### DIFF
--- a/src/Yasumi/Holiday.php
+++ b/src/Yasumi/Holiday.php
@@ -48,11 +48,6 @@ class Holiday extends DateTime implements JsonSerializable
     public const TYPE_OTHER = 'other';
 
     /**
-     * The default locale. Used for translations of holiday names and other text strings.
-     */
-    public const DEFAULT_LOCALE = 'en_US';
-
-    /**
      * @var array list of all defined locales
      */
     private static $locales = [];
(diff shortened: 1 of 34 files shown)


stelgenhof:
Thanks! This is indeed a good suggestion. I like it. 

Too bad you just missed the v2 release otherwise this could have been included. I think a v3 release is still far away :)

Anyway let's keep this PR, and as soon as a v3 is around the corner, it can be implemented.

Cheers! Sacha

c960657:
It was the release of Yasumi 2.0 that made me look into Yasumi again – hence the bad timing :-)

I have created a new PR, #123, that adds support for a configurable default locale in backwards compatible manner and deprecates passing a locale to `Yasumi::create()`. The actual removal of `$this->holiday` from `Holiday` and  `Provider` will have to wait for Yasumi 3.0.

PR #123, opened 2019-01-17
Allow default locale to be configured

c960657:
In preparation for the removal of `$this->holiday` from `Holiday` and `Provider` #122, this adds the suggested default language setting in backwards compatible way.

Two new class methods are added, `Yasumi::setDefaultLocale($locale)` and `Yasumi::getDefaultLocale()`, and a new optional argument is added to `$holiday->getName($locale = null)` that allows overriding the default.

This approach is used e.g. by the [PHP intl extension](http://php.net/manual/en/locale.setdefault.php), ~the [Symfony Translator](https://api.symfony.com/2.7/Symfony/Component/Translation/Translator.html)~, [Laravel](https://laravel.com/docs/5.7/localization), [Drupal's t() function](https://api.drupal.org/api/drupal/core%21includes%21bootstrap.inc/function/t/8.4.x) and the [Punic](https://punic.github.io/#Data-class) library.

Passing a locale to `Yasumi::create()` or `Yasumi:: createByISO3166_2()` is now deprecated (but still works as before). People can already now migrate to the new behaviour suggested for Yasumi 3.0 in #122. A deprecation warning is triggered following the tradition in [Symfony](https://symfony.com/doc/current/contributing/code/conventions.html#deprecations).

This paves the way for a second PR (#124) that adds support for fallback locales. I made two PR's in order to make code review easier. I recommend merging both PR's before doing a new release to ensure that the dust has settled on the API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - Observance holidays for Sweden [\#172](https://github.com/azuyalabs/yasumi/pull/172) ([c960657](https://github.com/c960657))
 - Added additional code style fixers and aligning StyleCI settings with PHP-CS.
 - Included extra requirement for some PHP Extensions in the composer file.
+- Added `Yasumi::setDefaultLocale()` and `Yasumi::getDefaultLocale()`. [\#123](https://github.com/azuyalabs/yasumi/pull/123) ([c960657](https://github.com/c960657))
+- Added optional $argument for `Holiday::getName($locale)` for overriding the default locale. [\#123](https://github.com/azuyalabs/yasumi/pull/123) ([c960657](https://github.com/c960657))
 
 ### Changed
 - Updated the translation for the All Saints holiday for the 'fr_FR' locale [\#152](https://github.com/azuyalabs/yasumi/pull/152) ([pioc92](https://github.com/pioc92))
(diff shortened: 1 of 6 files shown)


c960657:
@stelgenhof What do you think of this?

For the record: This does not break backwards compatibility but only deprecates some things. Such deprecations should preferably land early in the 2.x development cycle to allow users of the library to adjust their code already now, so their eventual upgrade to 3.x will be seamless.

c960657:
@Milamber33 The visibility changes are not part of this PR.

They were committed to the `develop` branch back in January 2019. 
326e8ecfd5a7162d0a999ebfbe2f86368cc56e08
22609ce0fa3a643fdeaa237d0834be3fd1538e44
d7b5e9274923bbb9bc3cd989bce350918224f327
13dd203fb9356cfaae58ce8f3bb994f2def38d0e
f1a4892c5cdebe058cf5f1fe5fa7e9024649d489

Milamber33:
@c960657 Okay, noted, I missed that. The question remains, though I grant that this is probably the wrong place for it now.

stelgenhof:
@Milamber33 As for the visibility of class members, you are right. When I started the project I didn't pay attention to the visibility much (I know: bad me!), however this should be changed of course.

stelgenhof:
@c960657 I had a look at the code of this PR and think it looks good. Just want to do a few more test rounds. 

My plan is do to a 2.2 release with this PR included. 2.2 will also contain all changes since v2.1 plus I am working on the Chile holiday provider. With this, also the documentation needs to be updated.

I don't have a particular target date yet in mind, but hopefully soon.

BTW: Big thanks for the coffee!! I feel I should return the favor as you have contributed much to Yasumi :)

Cheers! Sacha

c960657:
There is something about this that does not feel right. I seem to mix up default locale and fallback locale. I'll prepare an update. Stay tuned. I suggest this is merged after #176 (if that is accepted).

PR #124, opened 2019-01-17
Locale fallback

c960657:
This PR is a follow-up to #123.

This PR adds support for fallback locales. This has two major purposes:

1. The available translations are rather sparse, i.e. each provider only has full support for a few locales. When using a locale that is not fully supported, you may want to try a number of locales before falling back to English. It depends e.g. on whether your users are likely to understand the local language.
2. Allowing fallback from e.g. `en_US` to `en` introduces a kind of inheritance, so that strings shared between English variants are represented only once, while still allowing override for specific countries. This allows us to provide translations for all 102 `en_*` locales without repeating same same strings 102 times. This change makes the translation files less redundant, and we support a lot more locales.

Regarding #2: I removed duplicate entries for existing translations (e.g. if `en_US`, `en_GB` and all other variants were identical, I simply replaced them with one `en` entry). If there was a difference between e.g. `de_DE` and  `de_AT`, I generally chose the larger country as the general translation (i.e. the translation for `de_DE` becomes `de`). This is somewhat arbitrary, but it can easily be changed later on without changing the currently supported locales (it only affects locales that are _not_ currently supported).


c960657:
Finally got around to updating the translations inlined into the national providers.

c960657:
Closing in favour of #176 (too many merge conflicts and slightly different approach).

PR #126, opened 2019-01-19
Fix comment about weekend days in India

c960657:
Weekend in India is Sunday = 0. The code is right, the comment is wrong.


stelgenhof:
Thanks for catching this!

PR #133, opened 2019-02-07
International Womens Day

huehnerhose:
International Womens Day is a one time holiday in Germany/Berlin in 2019. 

Since it is a common holiday in Russia, I moved it to common holidays and added german localisation and the Germany/Berlin exception for 2019. 

I moved it to common holidays, because I think women are at least as important as animals. (World Animal Day seems only be used in the Netherlands but is in commonHolidays, I think International Womens Day deserves its place there, too)


stelgenhof:
Thanks for the PR! Any holiday can be added, as long as they are official/national or a defacto holiday in the respective provider (country) :) Just make sure you assign the correct type.

Cheers! Sacha

src/Yasumi/Provider/CommonHolidays.php
```diff
@@ -397,6 +397,41 @@ public function armisticeDay(
...
+     * @throws \Yasumi\Exception\InvalidDateException
+     * @throws \Yasumi\Exception\UnknownLocaleException
+     * @throws \InvalidArgumentException
+     * @throws \Exception
+     */
+    public function internationalWomensDay(
+        int $year,
+        string $timezone,
+        string $locale,
+        string $type = Holiday::TYPE_OFFICIAL
```
stelgenhof: I would suggest not to use TYPE_OFFICIAL. Official holidays are typically those days that governments have decided these are non-working days. I think for this case it is better to use TYPE_OBSERVANCE, or TYPE_OTHER.

huehnerhose: I'd like to disagree. Your definition for TYPE_OFFICIAL is 

    Type definition for Official (i.e. National/Federal) holidays.

And International Women's Day is a official holiday in several countries (see [Wikipedia](https://en.wikipedia.org/wiki/International_Women%27s_Day#Around_the_world) )

huehnerhose:
I updated Ukraine, since it also used a custom holiday add. 

Regarding the general unit test: I don't see where I should add this? There are tests for Russia, Ukraine and Germany/Berlin, which work now. Please give me a hint where I should add the general test. 

Thanks!

src/Yasumi/Provider/CommonHolidays.php
stelgenhof: I was speaking more generally as I didn't have any background information yet. Now I do :)
You are correct regarding the TYPE_OFFICIAL definition. However, I was considering other countries that don't celebrate this holiday yet and they may not be considering this as an official holiday. Anyway, that can be decided in the future :)

So, please leave it as official then.

stelgenhof:
@huehnerhose Thanks! Ignore my comment regarding the general unit test: I thought I always did that for common holidays, but that is not the case. Your unit tests are fine :)

Only one last request: Can you add an entry to the CHANGELOG.md file describing this change? That would be great. Once done, I'll have a final check and will merge it.

Cheers! Sacha

huehnerhose:
Sure :) and done.

PR #135, opened 2019-02-08
Fix composer warning [...]should not contain uppercase characters[...]

IceShack:
Deprecation warning: require-dev.mikey179/vfsStream is invalid, it should not contain uppercase characters. Please use mikey179/vfsstream instead. Make sure you fix this as Composer 2.0 will error.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
   "require-dev": {
     "friendsofphp/php-cs-fixer": "^2.14",
     "fzaninotto/faker": "~1.8",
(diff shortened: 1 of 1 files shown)


stelgenhof:
Thanks! Will change this as soon as Composer 2.0 has been officially released.

stelgenhof:
Didn't realize that composer >1.8.1 issues this warning as I hadn't upgraded my Composer yet :)

stelgenhof:
Made a messy merge, so in the end changed it myself.

Issue #132, opened 2019-02-07
Possibility to Set default timezone


IceShack:
All of my dates I'm using are in UTC, if I want to compare them with the holidays I always need to change the timezone.
So I think setting a default timezone either globally `Yasumi::defaultTimezone` or during creation as an extra parameter `Yasumi::create(Germany::class, 2017, 'de_DE', 'UTC')` would be very helpful.
As long as the default timezone is not set, it uses the timezone provided by the initialize method.
I'll create a PR once this is approved. :-D

stelgenhof:
The reason for the timezone not being configurable is that holidays should be calculated for the timezone (i.e. country) they are taking place in. If what you are suggesting would be done, then a strange situation occurs that for example Christmas may have already started (or finished) from the perspective of the other timezone.

Although that technically is correct, semantically has little meaning. You can consider for example New Year (Change at 00:00). If I would set New Year Midnight in Germany to a different timezone (for example Asia/Tokyo, which is UTC+9 hours), then the results would be that midnight has taken place already, where in reality it is still 9 hours before (UTC).

This is the primary reason, why a timezone isn't configurable, and any time conversation should be taken place outside this API. 

Hope it makes sense :)

Cheers! Sacha

IceShack:
I can't be the only one who stores dates as UTC in the database and relies on the frontend to decide which timezone the users in.
As I said before all I want is to make the timezone decision myself. As of now that is impossible, because everything happens in the `initialize` method and I can't change that value by Inheritance or even Reflection.
PS: I don't want to compare Germany New Years Eve with the Japanese one ;-)

stelgenhof:
Sure, I store my dates in UTC as well. I am not familiar with your business case, so let me show you what I do in one of my projects. 

I am sending a message to users when they login and that day is a public holiday. In this example I am using Japan's Mountain which takes place on August the 11th.

```
// User logged in on 2018-08-10 at 21:34:14 UTC (This is 2018-08-11 06:34:14 in Asia/Tokyo)
$login_date = Carbon::parse('2018-08-10 21:34:14', 'UTC'); // Date comes from the database

$japan = Yasumi::create('Japan', 2018);
$mountainDay = $holidays->getHoliday('mountainDay');

if($login_date->greaterThan($mountainDay)) {
    // Send the user an e-mail
}
```

The above example is a bit simplified of course and I am using the Carbon library for parsing and handling dates. In this case Carbon is taking care of the timezone conversions.

[...]

PR #136, opened 2019-02-08
Added getTimezone method

IceShack:
I think this would be a good compromise for #132 as it won't change the default business logic and won't break anything.


stelgenhof:
Closing this PR. See conversation in issue #132.

PR #145, opened 2019-05-17
Update pentecost.php

pioc92:
Add pentecost fr_FR translation


stelgenhof:
Thanks for the PR! I happen to have it merged already but I noticed there is no associated unit test for this holiday in France. Please create or update for each PR the appropriate unit test.

Cheers! Sacha

PR #146, opened 2019-05-17
Update easter.php

pioc92:
Add easter fr_FR translation


c960657:
FWIW, the change proposed in PR #124 would eliminate the need to create identical entries for multiple locales using the same base language.

PR #151, opened 2019-05-17
Update epiphany.php

pioc92:
Add epiphany fr_FR translation


pioc92:
It's not an official holiday, i close this PR

stelgenhof:
@pioc92 It is not a problem if a holiday is not official. You can still add it to Yasumi but then indicate it with a type other than OFFICIAL. It may be useful for other users to have this holiday even if it is not official.

PR #159, opened 2019-08-17
Simplify calculation for Sweden + Finland

c960657:
Slightly simpler/faster/more elegant calculation of All Saints Day and Midsummer Day in Finland and Sweden.


stelgenhof:
Thanks for the PR! Looking good; better than my initial code, haha.

PR #162, opened 2019-08-21
Add SubstituteHoliday

c960657:
I noticed that there is no translation for `substituteHoliday:secondChristmasDay` in the `UnitedKingdom` provider. I realised that providing translations for substitute holidays is cumbersome, especially when using global translations.

With this PR, substitute holidays are promoted into a first-class citizen with the `SubstituteHoliday` class. It references the substituted holiday and can inherit the translations from that with an “observed” suffix appended, if the substitute holiday itself does not have an original name.

The idea of using `{0}` as a placeholder is inherited from CLDR (see e.g. https://raw.githubusercontent.com/unicode-org/cldr/master/common/main/nl.xml).


src/Yasumi/data/translations/substituteHoliday.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @author Sacha Telgenhof <me@sachatelgenhof.com>
+ */
+
+// Translations for Whitsunday
+return [
+    'da_DK' => '{0} (erstatning)',
```
stelgenhof: Neat! I like this way :)

src/Yasumi/Filters/BankHolidaysFilter.php
```diff
@@ -43,12 +44,14 @@ public function accept(): bool
      */
     public function count(): int
     {
-        $days = \array_keys(\iterator_to_array($this));
+        $names = \array_map(static function (&$holiday) {
```
stelgenhof: This code block seems to occur in all the Filter classes. Do you think this could be refactored to an isolated code?

c960657: Moved to new method `Holiday::count()`. I'm not sure what the use-case is for this kind of counting, but I guess it is equally useful as a utility method.

c960657: We also use the counting in `AbstractProvider::count()`, so it is not specific to filters. But I could introduce an abstract filter class. Then at least it would only be duplicated twice. Do you prefer that?


--

> The use-case is straight forward in that each filter is able to return the total number filtered holidays. As substituted holidays don't need to be counted in the total number of holidays, [...]

This sounds like a very specific use-case (opinionated behaviour). For instance, in the following scenario, you would likely prefer a different definition:
```
print "Pay attention to these " . count($iterator) . " days:\n";
foreach ($iterator as $holiday) {
  print $holiday->format('Y-m-d') . ': ' . $holiday->getName() . "\n";
} 
```

I think it is rather surprising that `count($iterator)` does not reflect the number of `foreach` iterations.

But changing this would break BC, so it is not relevant for now. It was just to elaborate my comment about use-case :-)

stelgenhof: I was indeed thinking about an abstract filter class.

The `count($iterator)` correctly counts the number of iterations. The issue is actually when a holiday is being substituted, that holiday is counted twice (because a new Holiday instance is added to the internal array). 

The (domain) question is then: would you want to count the substituted holiday as well or should it be left out? At the time of building this, I felt a substituted holiday shouldn't be included in the count. :)

c960657: I have added an abstract filter class.

Let's discuss the domain question some other day :-)

PR #163, opened 2019-08-29
Suspended holidays in Portugal 2013-2015

c960657:
According to Wikipedia, four holidays were cancelled in three years; 2013, 2014 and 2015. Yasumi currently reports only 2014 and 2015.

Also, Portuguese Republic Day was also among the holidays. This is not currently supported by Yasumi.

From https://en.wikipedia.org/wiki/Public_holidays_in_Portugal#Revoked_holidays_in_2013–2015:
> In 2012, the Coalition government of Pedro Passos Coelho controversially revoked four holidays — two civilian holidays (Republic Day and Restoration of Independence) and two religious ones (Corpus Christi and All Saints Day). The move was effective from 2013 onwards [...] The four holidays were eventually restored by the government of António Costa, in January 2016.


stelgenhof:
Super! Thanks for this update.

PR #164, opened 2019-08-29
YAML-based tests

c960657:
I propose a new way to write tests for holiday providers.

Rather than testing each holiday separately, it may be easier to check all holidays in a given year for relevant years. This is particularly relevant for the current year +/- a few years, and for years in the past with various exceptions, such as the [suspended holidays in Portugal 2013-2015](https://github.com/azuyalabs/yasumi/pull/163), or substitute holidays, bridge holidays etc.

Listing all holidays for a given year makes it easy to check that all holidays (and no more) are returned for a given year, e.g. by comparing with other sources ([example](https://www.independent.co.uk/life-style/uk-bank-holidays-2019-what-days-off-get-how-many-easter-christmas-national-public-dates-a7533036.html)).

Also, some of the current tests tend to be a copy of the code they are testing ([example](https://github.com/azuyalabs/yasumi/blob/develop/tests/Ireland/StPatricksDayTest.php#L51)). This is fine for regression testing, but such tests are less likely to detect flaws in the logic during initial implementation.

For this PR I propose YAML for storing the expected test data. The YAML files are a bit verbose (though not as verbose as e.g. JSON), but we could consider crafting our own simple format for this purpose.

The YAML tests can be auto-generated, lowering the barrier for contributing new holiday providers. With the current approach to testing, it requires a substantial amount of work to provide test cases. The YAML files provided for this PR were all auto-generated (except the few `comment` entries). Of course, if you prefer TDD, you can also hand-edit the YAML files first.

The full lists of holidays makes it easy to help with QA, even if you are not familiar with the Yasumi code.

For this PR, I have added full coverage for at least one locale for all calendar providers for all years in the range 2010-2030. In addition to this, we should add years in the past of specific interest (exceptions, substitute holidays etc.). We could easily add more years – I don't have a strong opinion about this.

I have allowed the possibility to add comments to the YAML file, either for a specific holiday ([example](https://github.com/c960657/yasumi/blob/8d5b0fb3e0010b287450d7c8a6a0aa94deb3f188/tests/UnitedKingdom/en_GB.yaml#L335)), or the entire year (entries with no `date`and `name` entries but only a `comment` entry – [example](https://github.com/c960657/yasumi/blob/8d5b0fb3e0010b287450d7c8a6a0aa94deb3f188/tests/Portugal/pt_PT.yaml#L522)). This is a way to mention things to pay attention to, if the files are later changed (example).

If we adopt this approach, I think the YAML files can eliminate the need for some of the existing tests. However, I think the possible removal/adjustment of these should be a discussed in a separate PR.

For now I have just included all years 2010-2030. If you like this proposal, I will extend the YAML file with more relevant years for each specific provider.

What do you think?


c960657:
As an alternative to YAML, we could use Markdown. The source data is even more compact, and it is very review-friendly, because Github automatically formats Markdown.

Example:
```markdown
# 2012
Date       | Type     | Name                | Comment
---------- | -------- | ------------------- | ---
2012-01-02 | bank     | New Year's Day
2012-04-06 | official | Good Friday
2012-04-09 | bank     | Easter Monday
2012-05-07 | bank     | May Day Bank Holiday
2012-06-04 | bank     | Spring Bank Holiday | Moved to 4 June for the celebration of the Diamond Jubilee of Elizabeth II.
2012-08-27 | bank     | Summer Bank Holiday
2012-12-25 | official | Christmas Day
2012-12-26 | bank     | Boxing Day

# 2013
...
```

# 2012
Date       | Type     | Name    | Comment
---------- | -------- | ------- | ---
2012-01-02 | bank     | New Year's Day
2012-04-06 | official | Good Friday
2012-04-09 | bank     | Easter Monday
2012-05-07 | bank     | May Day Bank Holiday
2012-06-04 | bank     | Spring Bank Holiday | Moved to 4 June for the celebration of the Diamond Jubilee of Elizabeth II.
2012-08-27 | bank     | Summer Bank Holiday
2012-12-25 | official | Christmas Day
2012-12-26 | bank     | Boxing Day

Issue #99, opened 2018-03-05
Working days and other classification


c960657:
Working days

`AbstractProvider::isWorkingDay()` returns true for any defined holiday, including e.g. Valentines Day. This is surprising. I belive Valentines Day is a working day for everyone, even the most romantically inclined :-)

I guess all `TYPE_OFFICIAL` holidays are official non-workdays. But there are also some holidays `TYPE_OBSERVANCE` holidays that are de-facto non-workdays, at least for some companies/public offices.

For a large number of use-cases for showing holidays in business software, the only holidays of interest are those that affect the business, i.e. when other businesses (banks in particular) are closed. And not so much when people buy flowers to their sweethearts :-)

I know that “working day” is not a boolean property. But I think holidays can be tagged with different fairly objective properties that allow users to get a list of holidays that are relevant for a specific purpose.

Other use-cases include listing all bank holidays or holidays where shops are closed.

I suggest adding a new attribute in addition to the current `TYPE_xxx` property. This should be a bitmask or tag list that allows multiple flags/tags per holiday.

Currently a holiday can only have one type, so if a holiday is both a public holiday and a bank holiday, it can only be classified as one of them. This means that you cannot get a list of all bank holidays.

The list of flags could look something like this:
- Official non-working holiday.
- Other official or de facto non-working day, all companies/public offices are closed.
- Other official or de facto non-working day, a significant number of companies/public offices are closed.
- Official holiday or officially recognised day, working day.
- All banks are closed.
- All shops are closed.
- All schools are closed.
- Festive day celebrated by a significant part of the population.

The exact meaning of the words “all” and “significant” should be defined more clearly, e.g. >95% and >30%, respectively.

What do you think of adding such extra data?

stelgenhof:
Thank you very much for this! It is indeed not correct that `isWorkingDay` always returns true :(.

When I started with this library, I was already aware that the current types/categories wouldn't be sufficient or cover all situations. The current implementation is simple and would cover the majority of scenarios.

That said, I really like your proposal. Of course when introducing additional data, appropriate helper methods (API) should be present as well. To make it easier for people to use.

c960657:
This illustrates what I had in mind:
https://github.com/azuyalabs/yasumi/compare/develop...c960657:holiday-tags

I may have gone a bit overboard. On the other hand, as long as the tags are reasonably well-defined, it doesn't take a lot of work for a person with knowledge about a given country to add the relevant tags.

[...]

PR #169, opened 2019-09-03
Liberation Day is official in the Netherlands

c960657:
Liberation Day in the Netherlands is a public holiday, according to [Wikipedia](https://en.wikipedia.org/wiki/Public_holidays_in_the_Netherlands) and [Dutch law](https://wetten.overheid.nl/BWBR0002448/2010-10-10#Artikel3) (not mentioned by named but referred to as _fifth of May_).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - Fixed revoked holidays in Portugal in 2013-2015 [\#163](https://github.com/azuyalabs/yasumi/pull/163) ([c960657](https://github.com/c960657))
 - Fixed spelling issues in the Danish translation for Second Christmas Day. [\#167](https://github.com/azuyalabs/yasumi/pull/167)
 - Corpus Christi is official in Poland [\#168](https://github.com/azuyalabs/yasumi/pull/168) ([c960657](https://github.com/c960657))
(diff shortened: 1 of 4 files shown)


stelgenhof:
As the notes on Wikipedia said, May the 5th is not mandatory paid off for everybody. That's perhaps why initially I marked it as 'NON-OFFICIAL'. But now thinking about it may be better to mark it as OFFICIAL indeed.

c960657:
The distinction between the different holiday types is difficult due to different traditions in each country.

Many countries have legislation containing an explicit list of holidays. In my understanding, these have `TYPE_OFFICIAL`. Whether these days are paid/non-paid days off is a different matter – this often depends on tradition, collective agreements. With respect to non-working days, legislation often covers only the public sector.

Often the legislation mentions other days that are not official holidays but in some contexts have the same status. E.g. article 3.1 in the [Dutch law mentioned above](https://wetten.overheid.nl/BWBR0002448/2010-10-10#Artikel3) mentions Good Friday as such a day. These are probably `TYPE_OBSERVANCE`.

I assume the official classification is not so relevant for users of Yasumi. Instead I assume they are more interested in – depending on the specific use-case – whether it is a regular working day, whether shops are open, whether banks are open, etc. I am not sure how to model this, but I have outlined some thoughts in #99.

stelgenhof:
Yes correct. It may not always be easy to assign the proper indication. Furthermore the use-cases might be different for every user of Yasumi, so I tried to keep it as straightforward as possible (and think it should be that). For the moment this simple classification works but also understand and agree the proposal you made in #99 . I haven't yet properly thought about :(

PR #170, opened 2019-09-03
Change Dutch translation of Christmas Day

c960657:
I suggesting changing the Dutch translation for Christmas Day from _Kerstmis_ to _Eerste kerstdag_ for consistency with _Eerste paasdag_ and _Eerste pinksterdag_.


stelgenhof:
Although **Kerstmis** is correct, it is just the generic name for the festivity/event. So you're right; it is better to rename as you suggested.

PR #172, opened 2019-09-05
Swedish observance holidays

c960657:
In addition to official holidays, Sweden has several _de facto holidays_ and  _de facto half holidays_ (the term _de facto_ is used in the [English Wikipedia entry](https://en.wikipedia.org/wiki/Public_holidays_in_Sweden#De_facto_holidays_2) – I don't think a similar term is used in Swedish).

I think these de facto holidays should be marked as `TYPE_OBSERVANCE` (same argument as for Good Friday in the Netherlands, as mentioned in https://github.com/azuyalabs/yasumi/pull/169#issuecomment-527951090).

I am not Swedish, so I don't have a deep understanding of this. But I previously marked these days as “real” public holidays on my website and received several comments from Swedes these days are _not_ public holidays (_helgdager_).

This PR changes the type for Christmas Eve from `TYPE_OFFICIAL` to `TYPE_OBSERVANCE`, and adds the two other de facto holidays and the three de facto half holidays (all `TYPE_OBSERVANCE`).


stelgenhof:
I remember now that I was struggling how to categorize these. Perhaps OBSERVANCE is a better category indeed.

PR #175, opened 2019-09-12
Holiday provider for India

c960657:
New holiday provider with support for the [3 public holidays in India](https://en.wikipedia.org/wiki/Public_holidays_in_India).

I was not able to figure out when these days were made official holidays, but at least they are limited by the year of the event they are commemorating.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - Added additional code style fixers and aligning StyleCI settings with PHP-CS.
 - Included extra requirement for some PHP Extensions in the composer file.
 - Special subclass of Holiday for substitute holidays [\#162](https://github.com/azuyalabs/yasumi/pull/162) ([c960657](https://github.com/c960657))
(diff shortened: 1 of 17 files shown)


stelgenhof:
Nice! This only include the national holidays. Have you thought about the other, various religious holidays in India?

c960657:
I did think about it, and it is 100% relevant. But I'm afraid I don't have the time or need (or knowledge) to implement these.

There is a boatload of holidays for each of the large religions in India, so it is a considerable task. For my own purpose, I only need the official holidays, so I'll leave it there. Other developers are welcome to pick up this task and extend the provider.

If you don't want to merge this provider with only the official holidays, I totally understand.

stelgenhof:
Thanks! Totally understand about the volume. However usually my criterium for adding a new country is when all (or nearly all) official / recognized holidays are included. This to give end-users a complete package, rather then a half implemented one :)

Let me have a think about it. 

Thanks for the work anyway as always! Sacha

Milamber33:
Australia was far from complete before I fleshed it out...

stelgenhof:
@Milamber33 Exceptions can be made ;)

Anyways... although Australia wasn't immediately complete indeed, it was when it became part of a release.

PR #176, opened 2019-09-17
Locale fallback

c960657:
This PR is a new version of #124 (the previous version exploded in a giant conflict) that has been detangled from #123.

This PR adds support for fallback locales. This has two major purposes:

1. The available translations are rather sparse, i.e. each provider only has full support for a few locales. When using a locale that is not fully supported, you may want to try a number of locales before falling back to English. It depends e.g. on whether your users are likely to understand the local language.
2. Allowing fallback from e.g. `en_US` to `en` introduces a kind of inheritance, so that strings shared between English variants are represented only once, while still allowing override for specific countries. This allows us to provide translations for all 102 `en_*` locales without repeating same same strings 102 times. This change makes the translation files less redundant, and we support a lot more locales.

Regarding #2: I removed duplicate entries for existing translations (e.g. if `en_US`, `en_GB` and all other variants were identical, I simply replaced them with one `en` entry). If there was a difference between e.g. `de_DE` and  `de_AT`, I generally chose the larger country as the general translation (i.e. the translation for `de_DE` becomes `de`). This is somewhat arbitrary, but it can easily be changed later on without changing the currently supported locales (it only affects locales that are _not_ currently supported).


src/Yasumi/data/translations/substituteHoliday.php
```diff
@@ -12,11 +12,11 @@
 
 // Translation pattern for substitute holidays.
 return [
-    'da_DK' => '{0} (erstatning)',
-    'en_GB' => '{0} (substitute day)',
+    'da' => '{0} (erstatning)',
+    'en' => '{0} (substitute day)',
     'en_IE' => '{0} observed',
```
stelgenhof: Shouldn't these (en_IE, en_CA and en_US) be changed to the generic fallback locale as well?

stelgenhof:
Are you anticipating more changes to this PR? I'd like to review it and make it part of version 2.2

Cheers! Sacha

src/Yasumi/data/translations/substituteHoliday.php
c960657: Well, this change was actually intentional, but I may be wrong about the preferred translation for these days in various countries.

In UK, the term “Boxing Day (substitute day)” seems the preferred variant:
https://www.gov.uk/bank-holidays, https://www.thesun.co.uk/news/1736340/august-bank-holiday-uk-2019-public/

In the US, I believe the term “observed” is used. I guess this applies to Canada as well.

I don't know about Ireland. It was previously my impression that “observed” also was used there, but now I cannot find a good reference. So I have now reverted the override for en_IE, so it falls back to the generic “XXX (substitute day)”.

c960657:
I just pushed a few minor updates, but I don't anticipate any further changes.

src/Yasumi/data/translations/substituteHoliday.php
stelgenhof: Understood. Makes sense indeed. Anyway, in case needed we can always change :)

PR #177, opened 2019-10-01
Change http to https links

c960657:
HTTPS is the new black.

Update all links that currently redirect from HTTP to HTTPS. A few broken links were fixed.

This is a non-functional change, so I don't think it warrents a changelog entry (_less is more_).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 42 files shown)


stelgenhof:
I assume you've checked them one by one, right? ;)

c960657:
I checked that all affected hosts respond to HTTPS.

I just did an automated check for 404 errors and this revealed some more broken links. I have fixed those as well.

PR #178, opened 2019-10-04
Summer bank holiday test

c960657:
Fix the SummerBankHolidayTest for England, Wales and Northern Ireland which occasionally failed, because they failed to check for the renaming from August Bank Holiday to Summer Bank Holiday in 1965.

Also, update the tests for these countries to the coding standard introduced in 27c6e8173.


stelgenhof:
Thanks! Yes, also noticed this myself but hadn't yet time to fix it.

PR #182, opened 2019-10-24
Holiday providers for states of Austria

aprog:
Hello!

Thanks for the nice library. There are holidays that depend on state of Austria and this PR adds them.

### DIFF
--- a/src/Yasumi/Provider/Austria/Carinthia.php
+++ b/src/Yasumi/Provider/Austria/Carinthia.php
@@ -66,10 +66,10 @@ class Carinthia extends Austria
     private function calculatePlebisciteDay(): void
     {
         $this->addHoliday(new Holiday(
-          'plebisciteDay',
-          [],
-          new DateTime($this->year . '-10-10', new \DateTimeZone($this->timezone)),
-          $this->locale
+            'plebisciteDay',
+            [],
+            new DateTime($this->year . '-10-10', new \DateTimeZone($this->timezone)),
+            $this->locale
         ));
     }
 }
--- a/src/Yasumi/Provider/Austria/Carinthia.php
+++ b/src/Yasumi/Provider/Austria/Carinthia.php
@@ -65,6 +65,10 @@ class Carinthia extends Austria
      */
     private function calculatePlebisciteDay(): void
     {
+        if ($this->year < 1920) {
+          return;
+        }
+
         $this->addHoliday(new Holiday(
             'plebisciteDay',
             [],
--- a/src/Yasumi/Provider/Austria/Carinthia.php
+++ b/src/Yasumi/Provider/Austria/Carinthia.php
@@ -67,7 +67,7 @@ class Carinthia extends Austria
     {
       $this->addHoliday(new Holiday(
         'plebisciteDay',
-        ['en' => 'Plebiscite Day', 'de' => 'Tag der Volksabstimmung'],
+        [],
         new DateTime($this->year . '-10-10', new \DateTimeZone($this->timezone)),
         $this->locale
       ));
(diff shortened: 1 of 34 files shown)


stelgenhof:
Thanks for the PR! Very welcome indeed. Let me review it in the next few days.

src/Yasumi/Provider/Austria/LowerAustria.php
```diff
@@ -0,0 +1,75 @@
...
+     * November.
+     *
+     * @link https://en.wikipedia.org/wiki/Leopold_III,_Margrave_of_Austria
+     *
+     * @throws InvalidDateException
+     * @throws \InvalidArgumentException
+     * @throws UnknownLocaleException
+     * @throws \Exception
+     */
+    private function calculateStLeopoldsDay(): void
```
stelgenhof: Can you add a condition for when this holiday was first introduced/started? If not know, I would then use St. Leopold's birthday/day of death as a condition. (Guessing that before his birth this wasn't a holiday). 
This will prevent this holiday to be calculated before his birth.

src/Yasumi/Provider/Austria/Carinthia.php
```diff
@@ -0,0 +1,75 @@
...
+     * Kingdom of Serbs, Croats and Slovenes (Yugoslavia) after World War I.
+     *
+     * @link https://en.wikipedia.org/wiki/1920_Carinthian_plebiscite
+     *
+     * @throws InvalidDateException
+     * @throws \InvalidArgumentException
+     * @throws UnknownLocaleException
+     * @throws \Exception
+     */
+    private function calculatePlebisciteDay(): void
```
stelgenhof: Can you add a condition for when this holiday was first introduced/started?

src/Yasumi/Provider/Austria/Salzburg.php
```diff
@@ -0,0 +1,75 @@
...
+     * saint of the Austrian state of Salzburg.
+     *
+     * @link https://en.wikipedia.org/wiki/Rupert_of_Salzburg
+     *
+     * @throws InvalidDateException
+     * @throws \InvalidArgumentException
+     * @throws UnknownLocaleException
+     * @throws \Exception
+     */
+    private function calculateStRupertsDay(): void
```
stelgenhof: Can you add a condition for when this holiday was first introduced/started?

src/Yasumi/Provider/Austria/UpperAustria.php
```diff
@@ -0,0 +1,77 @@
...
+     * was to fight fires.
+     *
+     * @link https://en.wikipedia.org/wiki/Saint_Florian
+     *
+     * @throws InvalidDateException
+     * @throws \InvalidArgumentException
+     * @throws UnknownLocaleException
+     * @throws \Exception
+     */
+    private function calculateStFloriansDay(): void
```
stelgenhof: Can you add a condition for when this holiday was first introduced/started?

src/Yasumi/Provider/Austria/Vienna.php
```diff
@@ -0,0 +1,75 @@
...
+     * November.
+     *
+     * @link https://en.wikipedia.org/wiki/Leopold_III,_Margrave_of_Austria
+     *
+     * @throws InvalidDateException
+     * @throws \InvalidArgumentException
+     * @throws UnknownLocaleException
+     * @throws \Exception
+     */
+    private function calculateStLeopoldsDay(): void
```
stelgenhof: This method seems to be duplicate. You may want to create a Trait if more classes use the same method.

src/Yasumi/Provider/Austria/Salzburg.php
aprog: Looks like the holiday has been started on 710-9-24. Yasumi allows years from 1000. What is the best way to proceed?

stelgenhof: Ah. In that case you can leave it as is then :) Perhaps just add a note in the docblock.

aprog:
Thanks for review. Suggestions have been completed in the recent commits. :-)

PR #184, opened 2019-10-31
Locale override for getName()

### LINKED ISSUE #181: No easy way to get substituted holiday translations?
I love this package and have been using it for years. I used to be able to simply do ` $holiday->translations['en']` to get translations, even for substituted holidays. Now it seems like I have to do something like this monstrosity:
```
if ($holiday instanceof SubstituteHoliday) {
    $original = $holiday->substitutedHoliday;

    $holiday_en = $original->translations['en'];
    $holiday_jp = $original->translations['ja'];

    $substitute_en = $holiday->substituteHolidayTranslations['en'];
    $substitute_jp = $holiday->substituteHolidayTranslations['ja'];

    $en = str_replace('{0}', $holiday_en, $substitute_en);
    $jp = str_replace('{0}', $holiday_jp, $substitute_jp);
} else {
    $en = $holiday->translations['en'];
    $jp = $holiday->translations['ja'];
}
```

Is there no better way? Couldn't find it in the docs..

c960657:
Sometimes you need multiple translations for the same holiday. 

Pre v2.2.0 you could easily access all translations in `$holiday->translations`. With the introduction of locale fallback in #176, things got more complicated, especially with substitute holidays (introduced in #162), so accessing the `translations` property is difficult.

I suggest adding an optional `$locale` parameter to `$holiday->getName()`.

This fixes #181.

This suggestion is a subset of what was proposed in #123. This PR does not include the default locale part, only the locale override.

### DIFF
--- a/src/Yasumi/Holiday.php
+++ b/src/Yasumi/Holiday.php
@@ -157,10 +157,12 @@ class Holiday extends DateTime implements JsonSerializable
      * The name of this holiday is returned translated in the given locale. If for the given locale no translation is
      * defined, the name in the default locale ('en_US') is returned. In case there is no translation at all, the short
      * internal name is returned.
+     *
+     * @param string $locale the locale to use; if omitted, the display locale is used
      */
-    public function getName(): string
+    public function getName(string $locale = null): string
     {
-        foreach ($this->getLocales() as $locale) {
+        foreach ($this->getLocales($locale) as $locale) {
             if (isset($this->translations[$locale])) {
                 return $this->translations[$locale];
             }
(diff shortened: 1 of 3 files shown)


c960657:
One further possible feature would be allowing the user override the “default fallback locale chain”, e.g. by passing an array of locales instead of just a string.

E.g. `$holiday->getName('ca_ES')` would look for translations in Catalan with fallback to English, while `$holiday->getName(['ca_ES', 'es_ES', 'en'])` would look for Catalan, then Spanish, and finally English.

c960657:
I have created a alternative PR that implements the suggestion in my latest comment: #195.

stelgenhof:
@c960657 I have reviewed and merged your other PR. I guess we can close this one, right?

c960657:
Yes, this PR is no longer relevant.

PR #185, opened 2019-11-10
Do not capitalise names in DA/NB/NL

c960657:
In some languages, such as English, holiday names are considered proper nouns and must be capitalised. Other languages, such as Danish, treat them as common nouns (except _Kristi himmelfartsdag_ (Ascension Day) which includes the name of Christ), so they should not be capitalised when used mid-sentence. In Dutch, it is [a bit more complicated](https://onzetaal.nl/taaladvies/feestdagen/).

Yasumi currently provides capitalised names for all languages, except Swedish.

I suggest that all translations in Yasumi use the non-capitalised form, when applicable for the specific language. This supports most use cases.

If you need a capitalised name based on a non-capitalised version, you can just take the first letter and convert it into uppercase.

On the contrary, if you only have the capitalised version and need a version suitable for mid-sentence use, you cannot assume that you can just lowercase the first letter. Depending on the language and the specific holiday name in question, the first letter should or should not be lowercased (see e.g. the above mentioned complex rules for Dutch).

This PR covers Danish, Swedish, Norwegian and Dutch (based on my limited understanding of the rules for Dutch). Other languages may need a similar change, but this requires further investigation.

This change may be considered a BC break (despite the precedent for Swedish). If you think this is a serious concern, we could add a global configuration option — e.g. `Yasumi::capitalizeNames(bool $capitalize)` — which controls whether `getName()` should uppercase the first later of all names. The default should be true (for preserving BC). This should only be a temporary measure which is considered deprecated and will be removed in the next major version.


stelgenhof:
Thanks for the PR! Staying with the official / defacto spelling rules has my preference as well. As far as BC, not sure if this really is a BC change.

PR #188, opened 2019-12-11
Added fr translation for Second Christmas Day

Arkounay:


Arkounay:
> Can you add a unit test and update the CHANGELOG file accordingly?
> Thanks! Sacha


Hi,

The thing is that the day itself doesn't exist in France, it's only a translation (just like "Immaculate Conception" for instance, I believe it doesn't exist in english, it's not tested but there is the english key and value in `immaculateConception.php`). So I can't add a `SecondChristmasDayTest` that extends `FranceBaseTestCase`.

Is there any solution for this ?

I could also do another PR that adds a `LuxembourgProvider`, which does have the day and uses French as an official language so it would be testable in that case. 

Thanks

c960657:
I don't know much about French, but AFAICT _Lendemain de Noël_ is mostly used in Canada as a translation of Boxing Day. In Europe, this day is often referred to as _Saint Étienne_.

@Arkounay Any thoughts on this?

References:
* https://www.banquedeluxembourg.com/fr/bank/bl/contact_implantations_jours-feries
* [https://fr.wikipedia.org/wiki/Lendemain_de_Noël](https://fr.wikipedia.org/wiki/Lendemain_de_No%C3%ABl)
* [https://fr.wikipedia.org/wiki/Fête_de_la_Saint-Étienne
](https://fr.wikipedia.org/wiki/F%C3%AAte_de_la_Saint-%C3%89tienne)

Arkounay:
> I don't know much about French, but AFAICT _Lendemain de Noël_ is mostly used in Canada as a translation of Boxing Day. In Europe, this day is often referred to as _Saint Étienne_.
> 
> @Arkounay Any thoughts on this?
> 
> References:
> 
> * https://www.banquedeluxembourg.com/fr/bank/bl/contact_implantations_jours-feries
> * [https://fr.wikipedia.org/wiki/Lendemain_de_Noël](https://fr.wikipedia.org/wiki/Lendemain_de_No%C3%ABl)
> * [https://fr.wikipedia.org/wiki/Fête_de_la_Saint-Étienne
>   ](https://fr.wikipedia.org/wiki/F%C3%AAte_de_la_Saint-%C3%89tienne)

@c960657 

Looks like you're right.

I remember we spoke about this at work and picked "Lendemain de Noël" as a translation because the majority of us said that naturally and an app we use translated it as so. 

I think the app we use is based from this website, where it's translated as "Lendemain de Noël" for Luxembourg :
https://www.joursferies.fr/pays/luxembourg.php

However, if we check the homepage (https://www.joursferies.fr/) when it comes to France it's indeed translated as Saint-Étienne too there. Since these regions in France are close to Luxembourg, I thought that day had the same name.

So you're right, Saint-Étienne is the proper term in France.
I'm sorry for the mistake!

- For the `fr` locale we should definitely translate it as "**Saint-Étienne**"
- However for `fr_lu` I'd change it into "**Deuxième jour de Noël**" as it seems it's the official term there

I've checked official Luxembourg websites and there it's translated as translated as "Deuxième jour de Noël" :
https://luxembourg.public.lu/fr/vivre/qualite-de-vie/jours-feries-legaux.html 
https://itm.public.lu/fr/conditions-travail/conges/jours-feries.html
(I think these two websites are worth more than banquedeluxembourg.com since the `public.lu` domain is reserved for the government)

I can submit a new PR if you need

Thank you

c960657:
Thank you for your insights on this 👍 I went ahead and submitted a PR: #211.

PR #191, opened 2019-12-28
Fix issue if calculated workday is in next year

tniemann:
Currently, the year of the calculated date is not changed if it is in the next year.
When u try to get the next working day starting from 28.12.2019 and with 3 working days the calculated day is 1.1.2019 which is in the past.

### DIFF
--- a/src/Yasumi/Yasumi.php
+++ b/src/Yasumi/Yasumi.php
@@ -93,7 +93,7 @@ class Yasumi
 
         while ($workingDays > 0) {
             $date = $date->add(new \DateInterval('P1D'));
-            if (!$provider || (int)$date->format('Y') !== \getdate()['year']) {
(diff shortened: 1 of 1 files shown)


stelgenhof:
Thank you very much for the PR! Note that the checks failed as your branch is behind the current develop branch. You may want to do a rebase or merge.

tniemann:
Hey, sorry. I created a new PR.When do u think this will be in master and available via packagist? We use this library to calculate delivery dates which are all wrong at the moment!
Regards Thomas
    	
    	

    	
    
  




On Sat, Dec 28, 2019 at 2:03 PM +0100, "Sacha Telgenhof" <notifications@github.com> wrote:












Thank you very much for the PR! Note that the checks failed as your branch is behind the current develop branch. You may want to do a rebase or merge.

PR #195, opened 2020-01-03
Locale override for getName() (with multiple fallbacks)

c960657:
_This PR is an alternative to #184. This PR additionally adds support for multiple fallback locales._

Sometimes you need multiple translations for the same holiday. 

Pre v2.2.0 you could easily access all translations in `$holiday->translations`. With the introduction of locale fallback in #176, things got more complicated, especially with substitute holidays (introduced in #162), so accessing the `translations` property is difficult.

I suggest adding an optional `$locales` parameter to `$holiday->getName()`. This allows passing an array of locales to be searched for translations. E.g if `['es_ES', 'de_AT']` is provided, the following lookup priority is used: `es_ES` → `es` → `de_AT` → `de`.

If no parameter is provided, use the display locale with fallback to `DEFAULT_LOCALE` and the short name. E.g. if the display locale is `es_ES`, the following lookup priority is used: `es_ES` → `es` → `en_US` → `en` → _short name_.

Note that when providing the optional `$locales` parameter, `DEFAULT_LOCALE` and the short name is not used as fallbacks unless explicitly requested. This offers the user better control over fallback behaviour while preserving backwards compatibility.

Parent locales (e.g. `de` is the parent locale of `de_AT`) are always used, even when not specified explicitly. Parent locales are not considered a “fallback”. Instead, all strings in a parent locale are considered to be identical in the child locales unless explicitly overridden. I believe this convention is also used elsewhere, e.g. in CLDR.

### DIFF
--- a/src/Yasumi/Exception/MissingTranslationException.php
+++ b/src/Yasumi/Exception/MissingTranslationException.php
@@ -23,7 +23,7 @@ class MissingTranslationException extends BaseException implements Exception
      * Initializes the Exception instance
      *
      * @param string $shortName The short name (internal name) of the holiday
-     * @param string $locales   The locales that was searched
+     * @param array  $locales   The locales that was searched
      */
     public function __construct(string $shortName, array $locales)
     {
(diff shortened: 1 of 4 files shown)


stelgenhof:
Thanks again for the PR! Only had time for a quick look, but seems a good alternative.
Do you prefer this one to be included rather then the previous PR (#184) ?

c960657:
Yes, this is my favourite :-)

stelgenhof:
@c960657 Planning to do a minor release soonish, so I will review this PR this week.

PR #197, opened 2020-01-11
Use typographical apostrophe

c960657:
Proper typography prefers a  _curly/typographic apostrophe_ `’` rather than a _vertical/typewriter apostrophe_, `'` (see https://en.wikipedia.org/wiki/Apostrophe#Typographic_form). AFAICT this convention applies to all locales.

I suggest replacing all occurrences of typewriter apostrophe in translations with a typographical apostrophe.

Instead of editing all occurrences of Queen’s Birthday, I moved this to a separate translations file (and added a few translations based on Wikipedia).


stelgenhof:
Thanks! I'll guess I'll start to make some proper contributor documentation: there are quite a number of 'requirements' when creating providers or updates.

PR #201, opened 2020-01-15
Add substituted holidays for Australia

### LINKED ISSUE #199: Date addition (Australia Day Subdivision: AU-SA / Year: 2020)
It appears from https://www.safework.sa.gov.au/law-compliance/laws-regulations/public-holidays that in 2020 in South Australia, 2020-01-26 (Sunday) and 2020-01-27 (Monday) are both "Australia Day" public holidays. See specifically footnote 1 on this website which states: "If a Saturday, normal rates apply and the following Monday is the public holiday. If a Sunday, that and the following Monday are both public holidays."

Your project is currently only recording 2020-01-27 (Monday) as Australia Day.

It would be great if you could add the additional Sunday.

c960657:
Fixes #199.

I don't know anything about Australian holidays, but perhaps @jbroudou can verify this?

Basically we now mark both the original holiday and the substitute holiday as `TYPE_OFFICIAL` holidays. This applies to all holidays which are moved to the following Monday.


tests/Australia/AustraliaDayTest.php
```diff
@@ -75,22 +98,33 @@ public function testHolidayType(): void
...
+    {
+        return $this->generateRandomDates(1, 26, $this->timezone);
+    }
+
+    /**
+     * Returns a list of test dates
+     *
+     * @return array list of test dates for the holiday defined in this test
+     */
+    public function SubstituteHolidayDataProvider(): array
```
stelgenhof: The calculateHoliday function is not only used for 'Australia Day'. Ideally this test should also include assertions for other holidays that will get substituted.

jbroudou:
Hi, 

Is this applied to all Australian holidays? 

Not all holidays in Australia have the condition where if the public holiday falls on a Sunday, then both the Sunday and Monday are holidays. It depends on the holiday and state. 

Ideally you would set the BOTH condition (i.e. Sunday and Monday) on by default for all holidays, but then have a flag where you can switch it off so that the Monday is the only holiday.

c960657:
This PR does not change which holidays are moved to Mondays. Whether each holiday should be moved is specified by two boolean arguments for `calculateHoliday(..., $moveFromSaturday, $moveFromSunday, ...)`.

What this PR does is that _if_ a holiday is moved from a Saturday or Sunday to a Monday, the original holiday on Saturday/Sunday will be marked in addition to the holiday on Monday. The two days will be labeled e.g. “Australia Day” and “Australia Day observed”, respectively.

stelgenhof:
In addition to @c960657 comment, just wanted to confirm that the 'business' rules whether a holiday is being substituted lies already with each individual holiday. In some other countries, holiday substitution applies to all holidays so a more generic logic/code is applied.

stelgenhof:
@c960657 Would you mind having a look at my comments? Thanks!

c960657:
Sure, I'll take a look next week.

c960657:
Sorry for not responding to this sooner.

I have added an assertion on the holiday type for the subsituted holidays.

It is true that `calculateHoliday()` is used elsewhere, but only Australia Day uses the substitute holiday logic. Either it is called with `false, false`, meaning that no substitution should happen, or it is called with holidays that always fall on e.g. a Tuesday. This is why no other tests were failing after my changes. So I went ahead and replaced all these calls to `calculateHoliday()` with `addHoliday()`.

Also, I removed a bunch of unused asserts. If we just asserted that $foo is an instance of `Holiday`, we don't need to assert that it is not null.

stelgenhof:
@c960657 It's looking good, however can you again look at my comments? There could be an additional check for the SubstituteHoliday instance in the tests.

c960657:
I added assertSubstituteHoliday(). This asserts that the holiday is the right subclass. Doesn't this address your comment, or did you mean something else?

stelgenhof:
@c960657 Yes, thank you! I missed that the code was changed :) There are now some conflicts. Could you check these? Then I will merge the PR.

c960657:
Done :)

PR #205, opened 2020-02-18
Added Luxembourg Provider

Arkounay:


tests/Luxembourg/LuxembourgTest.php
```diff
@@ -0,0 +1,92 @@
...
+     * Tests if all official holidays in Luxembourg are defined by the provider class
+     * @throws ReflectionException
+     */
+    public function testOfficialHolidays(): void
+    {
+        $this->assertDefinedHolidays([
+            'newYearsDay',
+            'easterMonday',
+            'internationalWorkersDay',
+            'europeDay',
```
stelgenhof: Since Europe Day is celebrated on or after 2019, you need to account for that in this test. If you use test date before 2019 your tests will fail.

Arkounay: Thanks, I've changed the test, I think it's better this time ?

PR #220, opened 2020-06-07
Add accessors for shortName and substitutedHoliday

c960657:
Accessing public properties directly is bad practice and makes refactorings difficult.

This PR adds accessor methods for `$holiday->shortName` and `$substituteHoliday->substitutedHoliday` and deprecates direct access to those two properties.


stelgenhof:
That is definitely not the right way :) Thanks for the fix!

c960657:
Done :-)

PR #222, opened 2020-06-08
Feature/christmas eve

Patrick-Root:
Added christmas eve to germany as holiday::TYPE_OTHER.
It's not an official holiday, but in germany you get a half-day off.


stelgenhof:
Duplicate of #226

georgobermayr:
Hi,

Christmas Eve (24.12.) is a different day than New Years eve (31.12). So it would be great if this could be merged, since otherwise an important bank holiday is actually missing for Germany. 
Thank you!

PR #228, opened 2020-06-22
Fixing typo for Estonian Day of Restoration of Independence

reijovosu:
Source: https://et.wikipedia.org/wiki/Taasiseseisvumispäev


stelgenhof:
Thank you very much!

PR #231, opened 2020-09-27
Add non-working day flag

hktr92:
Hello.

I've developed a time tracking app back in 2017 that registers if an employee takes a period of time away from office. Because there are non-working days, we've implemented an entity that needs manual input from those data.

I've been looking for a replacement for a while, to get rid of manual input, then I found this project on Reddit. The integration in our project went flawless. Then I noticed some improvements that I could do to this project, in order to cut some lines of code when determining non-working days.

So I've forked this repository to add an `$nonWorkingDay` flag in `Holiday`, which is typed as `bool $nonWorkingDay = nul` and has a fallback to `$type === Holiday::TYPE_OFFICIAL` and added a filter class. The fallback is made in order to prevent BC, as most official holidays are non-working day (don't know about other countries). At least, this is the case in Romania, I'm don't know about other countries.

This is more like a shortcut to filter out official holidays, which from what I've read on the providers page, it might not be as reliable as it seems, so this little "hack" might do the trick.

Additionally, I've added a Dockerfile with php7.2-cli-alpine as I don't have any PHP runtime installed natively, with two helper scripts: `composer.sh` and `php.sh`. These aren't necessary to be merged with upstream, as I made these for my local development.

Don't know if the tests will pass, as I've ran `composer test` inside a docker container and it fails:
image

Hopefully, the code works without any issues, so it could be merged asap to upstream.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,37 @@
+FROM php:7.2-cli-alpine
+
+RUN apk add --update --no-cache \
+        git \
+        curl \
+        coreutils \
+        icu-dev \
+        libzip-dev \
+        freetype-dev \
+        libjpeg-turbo-dev \
+        libpng-dev \
+        bash
+
+RUN docker-php-source extract \
+    # ext-opache
+    && docker-php-ext-enable opcache \
+    # ext-calendar
+    && docker-php-ext-install calendar \
+    && docker-php-ext-enable calendar \
+    # ext-others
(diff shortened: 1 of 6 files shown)


stelgenhof:
Thanks for the PR! This is a hot-item at the moment (working day vs non-working day) and is something I am working on at the moment as well. Let me have a look at your PR in the next few days.

hktr92:
sure, take your time. i'll be available if needed.

PR #233, opened 2020-10-01
Improve Swiss providers

qligier:
This improves the Swiss providers by sticking to official holidays. A link to the local law has been added for the documentation.


src/Yasumi/Provider/Switzerland/Geneva.php
```diff
@@ -91,6 +92,7 @@ private function calculateJeuneGenevois(): void
         }
 
         $this->addHoliday(new Holiday('jeuneGenevois', [
+            'en' => 'Jeûne genevois',
```
sachatelgenhof, line 95: Not sure why you've added the English element just with the French translation. Is it because there is no English translation? In that case, you can leave out the English one as Yasumi will always pick the native language.

qligier: Indeed, it is unneeded. I've removed both English translations.

PR #235, opened 2020-11-19
Allow override WEEKEND_DATA in provider classes

mdhia:
Hello,

currently it is not possible to override the WEEKEND_DATA in a provider class. A use-case for this is, if you want to set Saturday as working day


mdhia:
Can someone take a look at this? 👀

stelgenhof:
Sorry for the delay, I will have a look ASAP.

stelgenhof:
In a future release this will be dealt with differently. For now this is fine I think.

PR #237, opened 2020-11-27
Add .gitattributes

fezfez:
This will drop downloaded package size drastically

before : 1749 items, size 4,2 MB
after    : 271   items, size 827,5 kB

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,12 @@
+/.github export-ignore
+/tests export-ignore
+/.gitattributes export-ignore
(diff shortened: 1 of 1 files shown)


stelgenhof:
Thanks! This is great. 
There are however some other test related files/folders (e.g. .github, .travis.yml) that can be excluded as well.|
Would you mind adding those as well?

fezfez:
Ok :+1:

fezfez:
@stelgenhof done and rebase ;)

PR #256, opened 2021-08-29
Canada provider: fix Canada Day definition and add National Day For Truth And Reconciliation

ovgray:
This PR fixes the definition of Canada Day in the Canada provider and adds National Day For Truth And Reconciliation

PR #257, opened 2021-09-01
update Canada Provider

ovgray:
correct Canada Day, add National Day For Truth and Reconciliation


ovgray:
this PR would makes the same changes to the 'develop' branch that #256 would make to the 'master' branch

PR #261, opened 2021-09-29
doing my part to remove any legacy of columbus

Jameron:
Indigenous Peoples' Day is finally recognized as a federal holiday in the USA.

### DIFF
--- a/src/Yasumi/Provider/USA.php
+++ b/src/Yasumi/Provider/USA.php
@@ -60,7 +60,7 @@ class USA extends AbstractProvider
         $this->calculateJuneteenth();
         $this->calculateIndependenceDay();
         $this->calculateLabourDay();
-        $this->calculateColumbusDay();
+        $this->calculateIndigenousPeoplesDay();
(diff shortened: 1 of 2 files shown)


alexparise:
This is an excellent and important change and I am fully in support of this change.

jlauters (approved):
Nice work!

alexparise (approved):
Good and important work.

stelgenhof:
@Jameron Thank you very much for your PR! Looking at the Wikipedia references, it seems that Indigenous Peoples' Day and Columbus Day are still both observed albeit in different states. Shouldn't Indigenous Peoples' Day than not be added as a holiday rather than replacing Columbus Day?

Jameron:
> @Jameron Thank you very much for your PR! Looking at the Wikipedia references, it seems that Indigenous Peoples' Day and Columbus Day are still both observed albeit in different states. Shouldn't Indigenous Peoples' Day than not be added as a holiday rather than replacing Columbus Day?

PR #264, opened 2021-11-17
Create Argentinian holidays provider.

nedSaf:


nedSaf:
## Update
- Added a new provider.
- Configured all the holidays and set all the descriptions from Wikipedia.

## ScreenShots
(This is a system we use for work logging, we pull the holidays into the work calendar)




## Next
- Add unit test for all the holidays in this provider. (Copying the tests from other providers)

## TB
### 3h

src/Yasumi/Provider/Argentina.php
```diff
@@ -0,0 +1,271 @@
...
+     * Initialize holidays for Argentina.
+     *
+     * @throws InvalidDateException
+     * @throws \InvalidArgumentException
+     * @throws UnknownLocaleException
+     * @throws \Exception
+     */
+    public function initialize(): void
+    {
+        $this->timezone = 'America/Santiago';
```
mariano-dagostino, line 47: @nedSaf `America/Argentina/Buenos_Aires`

mariano-dagostino (changes requested):
@nedSaf Checked all the days. They look good. The timezone is incorrect. Santiago is located in Chile.

src/Yasumi/Provider/Argentina.php
nedSaf: @mariano-dagostino Thanks :)

stelgenhof:
Thanks @nedSaf for your work! Nice to have a another country added

nedSaf:
## Update
- Added proper naming of the holidays also in English.
- Added tests for all the holidays.

## ScreenShots
#### Unit test result
Screen Shot 2021-11-19 at 11 52 55

### Holiday names in English


## Review
@stelgenhof This is ready for your review.

/cc @mariano-dagostino

nedSaf:
Fixing tests...

tests/Argentina/RemembranceDayTest.php
```diff
@@ -0,0 +1,88 @@
...
+     */
+    public const ESTABLISHMENT_YEAR = 2006;
+
+    /**
+     * Tests Day of Remembrance for Truth and Justice on or after 2006.
+     *
+     * @throws Exception
+     * @throws ReflectionException
+     */
+    public function testDiaDosFinadosAfter2006(): void
```
mariano-dagostino, line 45:
```suggestion
    public function testRemembranceDayAfter2006(): void
```

tests/Argentina/RemembranceDayTest.php
```diff
@@ -0,0 +1,88 @@
...
+            new DateTime("$year-03-24", new DateTimeZone(self::TIMEZONE))
+        );
+    }
+
+    /**
+     * Tests Day of Remembrance for Truth and Justice on or before 2006.
+     *
+     * @throws ReflectionException
+     */
+    public function testDiaDosFinadosBefore2006(): void
```
mariano-dagostino, line 61:
```suggestion
    public function testRemembranceDayBefore2006(): void
```

mariano-dagostino (approved):
@nedSaf looks good. There are two methods that could be renamed in the test.

stelgenhof:
@nedSaf You may want to check your tests as some of them are failing.

nedSaf:
@stelgenhof I fixed all the tests, ready for your review :)

src/Yasumi/Provider/Argentina.php
```diff
@@ -0,0 +1,307 @@
...
+            $easter = $this->calculateEaster($this->year, $this->timezone);
+
+            $carnavalMonday = clone $easter;
+            $this->addHoliday(new Holiday(
+              'carnavalMonday',
+              [
+                'en' => 'Carnival Monday',
+                'es' => 'Lunes de Carnaval',
+              ],
+              $carnavalMonday->sub(new DateInterval('P48D')),
```
stelgenhof, line 76: The sub method can return `false` on failure, whilst the constructor 3rd parameter can only takes a `DateTimeInterface` object. It would be better to handle this situation.

src/Yasumi/Provider/Argentina.php
```diff
@@ -0,0 +1,307 @@
...
+            ));
+
+            $carnavalTuesday = clone $easter;
+            $this->addHoliday(new Holiday(
+              'carnavalTuesday',
+              [
+                'en' => 'Carnival Tuesday',
+                'es' => 'Martes de Carnaval',
+              ],
+              $carnavalTuesday->sub(new DateInterval('P47D')),
```
stelgenhof, line 88: The sub method can return `false` on failure, whilst the constructor 3rd parameter can only takes a `DateTimeInterface` object. It would be better to handle this situation.

src/Yasumi/Provider/Argentina.php
```diff
@@ -0,0 +1,307 @@
...
+
+    /**
+     * Initialize holidays for Argentina.
+     *
+     * @throws InvalidDateException
+     * @throws \InvalidArgumentException
+     * @throws UnknownLocaleException
+     * @throws \Exception
+     */
+    public function initialize(): void
```
stelgenhof, line 45: This method is pretty fat. I recommend to extract the custom holidays logic into separate (private) methods to make the code more cleaner.

src/Yasumi/Provider/Argentina.php
```diff
@@ -0,0 +1,307 @@
...
+declare(strict_types=1);
+/*
+ * This file is part of the Yasumi package.
+ *
+ * Copyright (c) 2015 - 2021 AzuyaLabs
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @author Nader Safadi <nader@gizra.com>
```
stelgenhof, line 12: Please leave the original author name in this file comment. You can add your own name in the class comment.

tests/Argentina/ArgentinaTest.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Tests if all official holidays in Argentina are defined by the provider class.
+     *
+     * @throws ReflectionException
+     */
+    public function testOfficialHolidays(): void
+    {
+        $this->assertDefinedHolidays([
```
stelgenhof, line 46: This asserts that these holidays all are defined from the start test year (1980), which is correct. However it doesn't assert that the respective holidays are not present before their establishment year.

tests/Argentina/ArgentinaTest.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Tests if all observed holidays in Argentina are defined by the provider class.
+     *
+     * @throws ReflectionException
+     */
+    public function testObservedHolidays(): void
+    {
+        $this->assertDefinedHolidays([
```
stelgenhof, line 70: This asserts that these holidays all are defined from the start test year (1980), which is correct. However it doesn't assert that the respective holidays are not present before their establishment year.

stelgenhof (changes requested):

tests/Argentina/ArgentinaTest.php
nedSaf: But each test file has a method called `testNotHoliday`, which does exactly that (I know that they run for sure because they failed and I had to fix them), this structure of testing I copied from other providers so I think for consistency we should keep it as is.

But if you think otherwise, please let me know how you would like me to change the structure of the files?

tests/Argentina/ArgentinaTest.php
nedSaf: Same as above ^^

nedSaf:
@stelgenhof Fixed your comments, with one comment from me ^^

tests/Argentina/ArgentinaTest.php
stelgenhof: You're correct; I confused with some other edge cases I worked on a while ago (Sometimes I don't even remember my own code ;) )

src/Yasumi/Provider/Argentina.php
stelgenhof: I would make separate methods for each holiday, making it even more cleaner.

tests/Argentina/ArgentinaTest.php
stelgenhof: Just ran your test again and it fails:
```
1) Yasumi\tests\Argentina\ArgentinaTest::testObservedHolidays
Failed asserting that an array has the key 'carnavalMonday'.
```
You can replicate this by setting the starting test year lower than the establishment year. The error might not advertise itself as the actual test year is randomized.

stelgenhof (commented):
Please check my last comments.

src/Yasumi/Provider/Argentina.php
nedSaf: Done.

tests/Argentina/ArgentinaTest.php
nedSaf: @stelgenhof I'm not sure what are you asking me to do here.

Each test has its own `ESTABLISHMENT_YEAR` on which the whole test is based, this is what I found in the other providers.

> you can replicate this by setting the starting test year lower than the establishment year.

Yes, but it's set to a specific year on `setUp`, it will not fail unless you change the year.
The tests will fail if you change anything in the code, in this state, it will not fail.

Screen Shot 2021-11-24 at 9 13 22
Screen Shot 2021-11-24 at 9 13 39



In any case, please let me know exactly what you want me to do if you think I'm wrong, I'm not sure what I need to do.

stelgenhof: See for example this: https://github.com/azuyalabs/yasumi/blob/develop/tests/Georgia/GeorgiaTest.php#L65..L67

Your tests will not fail as your initial test start year (1980) is after any of the establishment years. It possibly can fail if you instantiate Yasumi with a year *before* any of the establishment years. Agreeably users might not do that a lot, but still is possible in case you'd like to see a backdated list of holidays.

nedSaf:
Pushed a fix.

nedSaf:
@stelgenhof Hey, I did push a fix after your comment, https://github.com/azuyalabs/yasumi/pull/264/commits/233f9d404e5f4f5f71eb20575805b5647dd4d0d0

I think that will cover the case you mentioned before, if not, please let me know.

nedSaf:
@stelgenhof any updates?

stelgenhof (approved):

Issue #270, opened 2021-12-23
United Kingdom - 2022 spring bank holiday


sylvaincombes:
In 2022, as we will have the Platinum Jubilee in United Kingdom on the 2022-06-03, it changes the Spring Bank Holiday date to the 2022-06-02

I will try to submit a PR about this

stelgenhof:
Closing. Resolved in v2.5

PR #271, opened 2021-12-23
fix: United Kingdom spring bank holiday 2022 (#270)

sylvaincombes:
United Kingdom spring bank holiday 2022 is moved to the 2nd june, due to the platinum jubilee

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ to [Semantic Versioning](https://semver.org).
 
 ### Added
 
+- United Kingdom spring bank holiday 2022 is moved to the 2nd june, due to the platinum jubilee [\#270](https://github.com/azuyalabs/yasumi/issues/270)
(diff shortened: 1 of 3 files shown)


stelgenhof:
Thanks for this PR! A little nitpicking: this is more of an enhancement and not a 'fix' since nothing is really broken :) 

According to this: https://en.wikipedia.org/wiki/Platinum_Jubilee_of_Elizabeth_II "the traditional Spring bank holiday will be moved from the end of May to the start of June". So some more changes are needed in this PR.

stelgenhof (commented):
Thanks for this PR! A little nitpicking: this is more of an enhancement and not a 'fix' since nothing is really broken :)

According to this: https://en.wikipedia.org/wiki/Platinum_Jubilee_of_Elizabeth_II "the traditional Spring bank holiday will be moved from the end of May to the start of June". So some more changes are needed in this PR.

stelgenhof:
Closing. Fixed in v2.5

PR #272, opened 2022-01-13
Add translation in English for German Unity Day

snikoletopoulos:
Add translation for German Unity Day in English. Please give me feedback.

### DIFF
--- a/src/Yasumi/Provider/Germany.php
+++ b/src/Yasumi/Provider/Germany.php
@@ -99,7 +99,11 @@ class Germany extends AbstractProvider
         if ($this->year >= 1990) {
             $this->addHoliday(new Holiday(
                 'germanUnityDay',
-                ['de' => 'Tag der Deutschen Einheit'],
+                [
+                    'en_US' => 'German Unity Day',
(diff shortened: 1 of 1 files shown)


c960657:
You don't need to specify `en_US` and `en_GB`  separately. Just specify `en` – this will work for all English variants.

snikoletopoulos:
Hello, thank you for your feedback

I tried using 'en' but I couldn't make it work. I noticed that in other translations there are 'en_US' etc. specified separately.
Can you direct me where to look to solve this?

stelgenhof:
@snikoletopoulos Could you share perhaps your code what you are trying to do? I used the following script and works fine using only the `en` locale.

```php
<?php

declare(strict_types=1);

require 'vendor/autoload.php';

$holidays = Yasumi\Yasumi::create('Germany', 2022);         

$unityDay = $holidays->getHoliday('germanUnityDay');

echo $unityDay->getName(['de']).PHP_EOL;
// prints "Tag der Deutchsen Einheit"     

echo $unityDay->getName().PHP_EOL;
// prints "German Unity Day"

PR #292, opened 2022-10-19
Suggest switching from getShortName to getName for reflection class to gain compatibility

SupraSmooth:
Firstly, thank you for this great library. Since this month is themed by the [Hacktoberfest](https://hacktoberfest.com/) a few months ago, I found a little bug working with this library, that I would like to fix.

We used a Symfony application with a lot of custom business logic and had the need to create a custom holiday provider to handle some application specific edge cases. Since this custom provider was located in our code, our namespace was obviously different from the one used in this library.

With this PR, I suggest switching to from `getShortName()` to `getName()` for the ReflectionClass created by the private method `anotherTime()` in the `AbstractProvider`.

We noticed that using `getShortName()` could result in a `ProviderNotFoundException` for some custom holiday providers, since the namespace is not fully qualified.

You can reproduce this bug by creating a custom holiday provider in a test project, where the namespace of the custom holiday provider differs in any way from the one used by this library. Next implement the `next()` or `previous()` methods from the `AbstractProvider` in your custom provider and see the application crash.

Greetings from Germany and keep contributing!


src/Yasumi/Provider/AbstractProvider.php
```diff
@@ -335,6 +335,6 @@ private function anotherTime(int $year, string $key): ?Holiday
         // Get calling class name
         $hReflectionClass = new \ReflectionClass(\get_class($this));
 
-        return Yasumi::create($hReflectionClass->getShortName(), $year, $this->locale)->getHoliday($key);
+        return Yasumi::create($hReflectionClass->getName(), $year, $this->locale)->getHoliday($key);
```
stelgenhof, line 338: The `getShortName` function was renamed to `getKey` in commit 829abcd971290a06c97a2c5e93e3bf1091cd713d. That should be the preferred function to be used here.

stelgenhof:
@SupraSmooth Thanks for spotting this! I added a review to your PR.

stelgenhof (changes requested):

src/Yasumi/Provider/AbstractProvider.php
SupraSmooth: In this PR I'm referring to the internal methods of the PHP reflection class object.
https://www.php.net/manual/en/reflectionclass.getname.php
https://www.php.net/manual/en/reflectionclass.getshortname.php

The renaming in the commit you mentioned does not affect this part of the project.
Sorry if I wasn't clear enough.

stelgenhof: Thanks for the explanation: I overlooked that it relates to the Provider classes, and mixed up the `getName` and `getShortName` functions in the `Holiday` class.

stelgenhof (approved):

PR #297, opened 2022-12-19
[fix-296] make epiphany an official holiday in German states, fixes #296

### LINKED ISSUE #296: German holiday epiphany should be an official holiday
The Holiday on the 6th of January should be an official holiday in Baden-Württemberg, Bavaria and SaxonyAnhalt.

Currently, it is marked as type "other" instead of "official"

AnnaDamm:
see #296


stelgenhof:
Thank you for the PR! I'll have a look.

stelgenhof (approved):

PR #306, opened 2023-01-06
Adds ext-intl to required extensions

freshleafmedia:
The `ext-intl` extension is required to run tests since 111c2e6931b50d758923ca883aebbc15d83d29f4. Specifically due the to usage of `intltz_get_tz_data_version`

### DIFF
--- a/.github/workflows/coding-standard.yml
+++ b/.github/workflows/coding-standard.yml
@@ -30,6 +30,7 @@ jobs:
         with:
           php-version: ${{ matrix.php-versions }}
           coverage: pcov
+          extensions: intl
 
       - name: Get Composer Cache Directory
         id: composer-cache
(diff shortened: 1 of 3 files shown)


stelgenhof:
Thank you @freshleafmedia for this PR. I am aware of this and had not yet added the extension as a requirement since the CI pipeline is failing for the Windows OS. I recall trying to add the extension in the GitHub actions but didn't seem to solve the issue. I left it since, but should look into it again.

freshleafmedia:
That's odd. The tests were passing before so that must mean that the extension is present. Otherwise an exception would have been thrown :thinking:

It's as though Composer doesn't detect it but it is there at runtime? Is it using a different `php.ini` maybe? Doesn't make any sense to me

stelgenhof:
What OS are you using locally then? As you can see here in the actions log https://github.com/azuyalabs/yasumi/actions/runs/3854998767/jobs/6569582950#step:8:19 the extension seems to be missing.

PR #311, opened 2023-03-06
Add International Womens Day to DE-MV

ihmels:


stelgenhof (changes requested):
Can you add the appropriate unit tests for your change? Also, if you have any documentation or reference for this change, that would be great.

stelgenhof:
@ihmels Can you add the appropriate unit tests for your change? Also, if you have any documentation or reference for this change, that would be great.

ihmels:
I have added a test.

The source is https://www.ndr.de/nachrichten/mecklenburg-vorpommern/Frauentag-in-MV-Landtag-beschliesst-neuen-Feiertag,frauentag370.html. However, there is [another source](https://en.wikipedia.org/wiki/Public_holidays_in_Germany) on Wikipedia that is already referenced via `Germany::getSources()`.

Where is the best place to put the source?

stelgenhof:
@ihmels Good point regarding the source. You could override the `getSources` method while adding the additional link.

ihmels:
@ stelgenhof I've added the source and fixed the namespace in the test. Can you run the workflow again for these changes?

src/Yasumi/ProviderInterface.php
```diff
@@ -34,7 +34,7 @@ public function initialize(): void;
      * Returns a list of sources (i.e. references to websites, books, scientific papers, etc.) that are
      * used for determining the calculation logic of the providers' holidays.
      *
-     * @return array<string> a list of external sources (empty when no sources are defined)
+     * @return list<string> a list of external sources (empty when no sources are defined)
```
stelgenhof, line 37: I am not familiar using 'list' as the return type. Are you using a particular tool that requires this to be 'list' instead of 'array'. Furthermore, this change causes a warning by Phan: 

`The inherited return type 'list<string>' for Yasumi\ProviderInterface::getSources is more specific than the implemented return type for Yasumi\Provider\Argentina::getsources 'array<array-key, string>' `

ihmels:
@stelgenhof Can you approve the workflow again?

src/Yasumi/ProviderInterface.php
stelgenhof: Apologies for the late reply. I am just a bit reluctant to use tool specific syntax/conventions.

ihmels:
> @ihmels Can you check you once more?

The error was.

> ```
>ERROR: InvalidReturnStatement - src/Yasumi/Yasumi.php:245:16 - The inferred type 'array<string, array<array-key, string>|null|string>' does not match the declared return type 'array<string, string>' for Yasumi\Yasumi::getProviders (see https://psalm.dev/128)
>        return $providers
>```

This error was not related to my changes, but I fixed it. Can you run tests again?

src/Yasumi/ProviderInterface.php
ihmels: I don't think that this syntax is specific for Psalm anymore. Also PHPStan and PhpStorm support this syntax.

stelgenhof:
@ihmels I've made some fixes on the `develop` branch that would make your PR error free. Can you rebase your branch? Thanks.

stelgenhof (approved):

PR #315, opened 2023-06-09
Adjust naming of Australian public holiday from Queen's Birthday to King's Birthday

FuzzyWuzzyFraggle:


stelgenhof:
@FuzzyWuzzyFraggle Thank you very much for the PR. I think it is better to make the name changes time dependent, i.e. being able to see this name change as part of its timeline. Since it is an official event we want to track that in in time.

FuzzyWuzzyFraggle:
@stelgenhof are you proposing that we leave the current implementation of the Queen's birthday intact, and add in a finish date, and create a brand new holiday for the King's birthday with a start date?

The underlying legislation in most of the jurisdictions that this holiday is observed have it as a generic holiday (admittedly they rever to it as the Sovereigns birthday, not the Monarchs birthday) and the update is purely a naming change

https://legislation.nsw.gov.au/view/html/compare/2012-07-06/2023-01-13/act-2010-115
https://www.legislation.qld.gov.au/view/whole/html/inforce/2017-06-05/act-1983-018
https://content.legislation.vic.gov.au/sites/default/files/ad24ad2c-06f2-3ae3-b0ce-1fcd9b5f61a0_93-119aa026%20authorised.pdf
https://www.legislation.nt.gov.au/en/Legislation/PUBLIC-HOLIDAYS-ACT-1981

stelgenhof:
@FuzzyWuzzyFraggle Yes indeed. That would allow to go back in time. I am aware that many using Yasumi have no such need and in this case it's only a name change. However, I know there are use cases where you would like to go back in time to get the list of holidays as it was then.

PR #317, opened 2023-06-20
Added Nigeria Holiday

mitmelon:
Added Nigeria Holiday

### DIFF
--- a/src/Yasumi/Provider/Nigeria.php
+++ b/src/Yasumi/Provider/Nigeria.php
@@ -1,0 +1,104 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Yasumi package.
+ *
+ * Copyright (c) 2015 - 2023 Manomite
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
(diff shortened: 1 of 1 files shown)


stelgenhof:
@mitmelon Thank you very much for your PR, however you have not added any unit-tests. Without unit-tests makes it hard to verify your code is working (and as designed). Please follow the guidelines outlined here: https://github.com/azuyalabs/yasumi/blob/develop/CONTRIBUTING.md

Also you have two PRs (#316 and #317) that contain the same file: this will cause conflicts when merging. It is better to keep everything in a single PR.

mitmelon:
> @mitmelon Thank you very much for your PR, however you have not added any unit-tests. Without unit-tests makes it hard to verify your code is working (and as designed). Please follow the guidelines outlined here: https://github.com/azuyalabs/yasumi/blob/develop/CONTRIBUTING.md
> 
> Also you have two PRs (#316 and #317) that contain the same file: this will cause conflicts when merging. It is better to keep everything in a single PR.

Okay... Thanks... Will correct all this tomorrow and add testcase...as stated.

stelgenhof:
@mitmelon Just wanted to follow up after my last reply. Do you still need any help with your PR?

PR #318, opened 2023-06-23
Adding missing countries.

miguelanruiz:
Hey there,

Now is missing Colombia, Peru and Mexico, so i think you could include with this pullr, main idea is another applications can has updated dates.

Best regards,

Issue #143, opened 2019-03-30
Provider for Mexico


gogl92:
Creating the provider for Mexico is not too difficult except for the holidays that are adjusted to the closest monday.

stelgenhof:
Thanks! Is this issue just a request or are you planning on contributing the provider for Mexico?

Having holidays that are observed to the next Monday in case a national holiday is not difficult. You can see the code of other holiday providers how it is done.

Cheers! Sacha

gogl92:
I will try to make the contribution and create the pull request.

stelgenhof:
@gogl92 Cool! Let me know if you have any questions. Best is to use one of the existing countries as a starting point. FIY, your PR needs to be made against the develop branch :)

Cheers! Sacha

stelgenhof:
@gogl92 Wanted to check if you have made any progress with the Mexico provider. If not and you don’t have any time, I can take it over if you like

PR #329, opened 2023-12-04
Mexico Provider

gogl92:
Time ago I created a ticket to implement Mexico's provider but I didn't have the time to implement it #143 

I saw this will conflict with #318 but looks like the PR it's stale at the moment, but I'm open to see if we can move that one instead or merge this and update the PR later, please let me know your thoughts @stelgenhof


stelgenhof:
@gogl92 Thank you very much for this PR! Yes, the Mexico provider is indeed also included in #318 but - as you said - I haven't had any feedback on that PR. So let's incorporate yours!

gogl92:
> All looking good!

Thank you! Having the pipelines running in my fork made it really easy!

PR #332, opened 2024-01-15
Update methods visibility in multiple Providers

attepulkkinen:
Methods in several providers including Switzerland, South Korea, and Luxembourg among others have been updated from private to protected. These changes allow for better extensibility and customization of the Yasumi library. The methods affected are those calculating specific holidays in the different regions.


stelgenhof (changes requested):
The changes look good, except you have some unresolved code style issues. You can fix  these by running the `composer format` command. After that, this PR should be good to go :)

attepulkkinen:
Running that command on my computer gives no errors, but I'll write a fix manually.

Also, not sure if this is a known issue, but there is also the following deprecation warning

```
Detected deprecations in use:
- Rule set "@PER" is deprecated. Use "@PER-CS" instead.
```

attepulkkinen:
Actually, this is a bit silly style fix. I don't think such type `Australia\string` actually exists…


```diff
--- /home/runner/work/yasumi/yasumi/src/Yasumi/Provider/Australia/AustralianCapitalTerritory.php
+++ /home/runner/work/yasumi/yasumi/src/Yasumi/Provider/Australia/AustralianCapitalTerritory.php
@@ -61,11 +61,11 @@
      *
      * @see https://en.wikipedia.org/wiki/Easter
      *
-     * @param int         $year     the year for which Easter Saturday need to be created
-     * @param string      $timezone the timezone in which Easter Saturday is celebrated
-     * @param string      $locale   the locale for which Easter Saturday need to be displayed in
-     * @param string|null $type     The type of holiday. Use the following constants: TYPE_OFFICIAL, TYPE_OBSERVANCE,
-     *                              TYPE_SEASON, TYPE_BANK or TYPE_OTHER. By default an official holiday is considered.
+     * @param int                   $year     the year for which Easter Saturday need to be created
+     * @param string                $timezone the timezone in which Easter Saturday is celebrated
+     * @param string                $locale   the locale for which Easter Saturday need to be displayed in
+     * @param Australia\string|null $type     The type of holiday. Use the following constants: TYPE_OFFICIAL, TYPE_OBSERVANCE,
+     *                                        TYPE_SEASON, TYPE_BANK or TYPE_OTHER. By default an official holiday is considered.
      *
      * @throws \Exception
      */
```

The long-term solution here would be to use enums when support for PHP 8.0 will be eventually dropped.

stelgenhof:
Thanks for noticing the deprecated ruleset. With recent PHP CS fixer updates I have seen some similar odd changes (and reverted). 

> Actually, this is a bit silly style fix. I don't think such type `Australia\string` actually exists… How do you want to continue?

Yes this indeed is a bit silly.  Let me see if I can correct all of this.

stelgenhof:
@attepulkkinen I've made a small (hopefully temporary) change that addresses the latest PHP CS Fixer issues. You may want to rebase your branch and give it another try.

stelgenhof (approved):

attepulkkinen:
Thank you. Can you tell when the next version will be released?

stelgenhof:
@attepulkkinen Apologies for missing your last question. Generally I release a new version if there are a number of valuable changes since the last release. So, at the moment, I don't have any new release scheduled soon.

PR #334, opened 2024-04-02
Fix, Provider\Germany: Pentecost is not an official holiday - except in Brandenburg

fbett:
Hi,

Pentecost was originally removed from the list of general holidays in #100.
That's because pentecost is not an official holiday (until today) - except in Brandenburg.

I don't know why, it was added again in #225.

So i don't remove pentecost from the list of general holidays again, but it is generally marked as TYPE_OTHER.


fbett:
Hi @stelgenhof, Thanks a lot for your feedback. You are right, there was another test i missed. It's updated now. 
I raised the memory_limit on my local machine to run all the tests.

stelgenhof:
@fbett I have fixed the code styling issues. Please rebase your branch. If all goes well, we can then merge your PR :)

PR #337, opened 2024-04-10
Fix, Provider\Germany: Pentecost is not an official holiday - except in Brandenburg

fbett:
Hi,

Pentecost was originally removed from the list of general holidays in https://github.com/azuyalabs/yasumi/pull/100.
That's because pentecost is not an official holiday (until today) - except in Brandenburg.

I don't know why, it was added again in https://github.com/azuyalabs/yasumi/pull/225.

So i don't remove pentecost from the list of general holidays again, but it is generally marked as TYPE_OTHER.

Previous #334


stelgenhof:
@fbett Your code is failing because of new PHP CS rules. Please run the following on your PR/branch:
`composer cs-fix`. And also run the tests locally first before committing: `composer test`.

When all is without errors, please commit. Then it should be good! :)

fbett:
The PR is refreshed.

stelgenhof (approved):

PR #340, opened 2024-05-29
Update Slovakia.php

hamrak:
One-time holiday was added on 30.11.2018 and removed 1.9. since 2024, because our Slovak government has officially decided that we don't need so many days to rest 🤦🏼


stelgenhof:
@hamrak Thank you for the PR. You may want to update the unit tests to reflect the updates you made. Also the Coding Standards check is failing so you need to fix those as well. Thanks!

stelgenhof (approved):
Thanks! All looking good!

PR #343, opened 2024-08-19
handle invalid timezones Australia/ACT and Europe/Kiev

### LINKED ISSUE #342: BUG: Unknown or bad timezone Australia/ACT
Hi and thanks for Yasumi!

I ran into an issue: the used timezone `Australia/ACT` is not an official timezone in PHP, at least in the latest `timezone database version 2024.1`:
- https://www.php.net/manual/en/timezones.australia.php 
- https://www.php.net/manual/en/timezones.php

Log entry:
```
CRITICAL: Uncaught PHP Exception DateInvalidTimeZoneException: "DateTimeZone::__construct(): Unknown or bad timezone (Australia/ACT)" at DateTimeZoneFactory.php line 31 {"exception":"[object] (DateInvalidTimeZoneException(code: 0): DateTimeZone::__construct(): Unknown or bad timezone (Australia/ACT) at /var/www/kimai/vendor/azuyalabs/yasumi/src/Yasumi/Provider/DateTimeZoneFactory.php:31)"} []
```

It happened with PHP 8.3.6.

This comes from the [AustralianCapitalTerritory](https://github.com/azuyalabs/yasumi/blob/develop/src/Yasumi/Provider/Australia/AustralianCapitalTerritory.php#L36).

kevinpapst:
Fixes #342

I hope the code comments explain the reasoning behind the changes.

All my systems seem to support both variants, but I had multiple bugs raised about `Australia/ACT` and `Europe/Kiev`.
Can't say when this happens, maybe it depends on the `tzdata` lib on the system where PHP was compiled. Or some PHP builds remove all timezone name which are only so called "links". 

In the official table (see https://en.wikipedia.org/wiki/List_of_tz_database_time_zones ) both problematic TZ are only links, so I replaced them with their real TZ identifier.

Edit: I used two approaches, so you can decide which way is the better one. 
Use the correct TZ identifier directly (Sydney) or use the try&catch approach (Kiev). 
Maybe try&catch in both places is the best approach for maximum BC?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ changes.
 
 ### Changed
 
+- Holiday calculation methods in providers are now protected instead of private 
+  to allow use in [custom providers](https://www.yasumi.dev/docs/cookbook/custom_provider/).
+  [\#331](https://github.com/azuyalabs/yasumi/issues/331)
+
 ### Fixed
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,14 @@ changes.
 
 ### Changed
 
-- Holiday calculation methods in providers are now protected instead of private 
+- Holiday calculation methods in providers are now protected instead of private
   to allow use in [custom providers](https://www.yasumi.dev/docs/cookbook/custom_provider/).
   [\#331](https://github.com/azuyalabs/yasumi/issues/331)
 
 ### Fixed
+
+- Handle invalid/changed timezones Australia/ACT and Europe/Kiev [\#343](https://github.com/azuyalabs/yasumi/pull/343)
+  ([Kevin Papst](https://github.com/kevinpapst))
 
 ### Removed
 
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -1,9 +1,11 @@
 <?php
 
-declare(strict_types=1);
+declare(strict_types = 1);
 
 /**
- * This file is part of the Yasumi package.
+ * This file is part of the 'Yasumi' package.
+ *
+ * The easy PHP Library for calculating holidays.
  *
  * Copyright (c) 2015 - 2024 AzuyaLabs
  *
@@ -13,7 +15,13 @@ declare(strict_types=1);
  * @author Sacha Telgenhof <me at sachatelgenhof dot com>
  */
 
-$config = new AzuyaLabs\PhpCsFixerConfig\Config();
+$config = new AzuyaLabs\PhpCsFixerConfig\Config('2015', null, 'Yasumi');
 $config->getFinder()->in(__DIR__)->notPath('var');
+
+$defaults = $config->getRules();
+
+$config->setRules(array_merge($defaults, [
+    'php_unit_method_casing' => ['case' => 'camel_case'],
+]));
 
 return $config;
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -12,29 +12,8 @@ declare(strict_types=1);
  *
  * @author Sacha Telgenhof <me at sachatelgenhof dot com>
  */
-$finder = PhpCsFixer\Finder::create()->in(__DIR__);
 
-$config = new PhpCsFixer\Config();
-$config->setRiskyAllowed(true)->setRules([
-  '@PER' => true,
-  '@Symfony' => true,
-  'combine_consecutive_issets' => true,
-  'combine_consecutive_unsets' => true,
-  'explicit_string_variable' => true,
-  'no_superfluous_elseif' => true,
-  'no_superfluous_phpdoc_tags' => ['remove_inheritdoc' => true],
-  'not_operator_with_successor_space' => true,
-  'nullable_type_declaration_for_default_null_value' => ['use_nullable_type_declaration' => true ],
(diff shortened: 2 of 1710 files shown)


stelgenhof:
Thank you very much @kevinpapst for this PR! Unfortunately there are sometimes backwards incompatible changes introduced in PHP. In the past there were similar BC issues with historical DST changes between PHP versions.

We could also move away from the timezone identifiers (type 3) altogether, and use the UTC offset (type 1) or the timezone abbreviation (type 2), as these seem less prone to changing. The benefit of the timezone identifier is of course it is more human friendlier than the other types.

kevinpapst:
Yeah, I fought in the past already with the `Kiev` removal, but that was in the scope of my time-tracking app, so I could take care of it myself. I use try&catch for it.

I think the timezone identifier is totally fine to use, it doesn't change regularly and the try&catch approach works flawless to handle this situation. Changing it to TZ offset would be a bigger BC break, as many of the `$timezone` variables are public and therefor you would change the public API. 

How to move forward?
1. Shall I change the the Australia class to try&catch as well? 
2. Never used psalm, what is the equivalent of `@phpstan-ignore-line` for the "dead catch" warning?

stelgenhof:
Yeah, wouldn't think the timezone identifier changes regularly either, but thought the others could be an option instead. However indeed that would break Yasumi's API :( 

I think the try&catch approach is a good approach. Please go ahead and use the same for Australia. Let me check on that dead catch issue in regards to Psalm.

kevinpapst:
I already did change Australia. 
And Psalm doesn't report it, it was only PHPStan. 
Checks are green, ready for review/merge from my end.

stelgenhof (changes requested):
Looking good! Could add unit tests for this and update the Changelog as well?
Thanks!

kevinpapst:
Ok for the Changelog, but tell me how to test these changes.

The result depends on the environment / PHP runtime.

thrashzone13:
Hi @stelgenhof  @kevinpapst !

I just wanted to suggest using [phpversion()](https://www.php.net/manual/en/function.phpversion.php) . It's self explaining and better than using a try catch

kevinpapst:
Does not depend on phpversion afaik, so this approach is not usable. 
Afaik this depends on the libicu version PHP was compiled against (or something like that, not sure about the exact internals). But the issue happened with PHP 8.1 and 8.3 and I have enough systems with these version where it doesn't happen.

stelgenhof:
I was thinking the same to use the PHP version for testing. If that turns out to be difficult, we can forgo the tests :)

stelgenhof:
@kevinpapst Can you only update the changelog? If done, then I'll merge it. Thanks!

kevinpapst:
@stelgenhof Changelog adjusted: done. 

Any chance you could create a bugfix release? I have production bugs with several users.

CHANGELOG.md
```diff
@@ -14,12 +14,15 @@ changes.
 
 ### Changed
 
-- Holiday calculation methods in providers are now protected instead of private 
+- Holiday calculation methods in providers are now protected instead of private
```
kevinpapst, line 17: sorry, that must have been my IDE 🤷

kevinpapst:
Hello @stelgenhof ,

I keep on having users who run into this bug. I understand if you don't want to merge this: its open source, your product and there is no warranty given. But I am unfortunately getting bug/support requests every other week and I have no proper answer since a few months. 

I don't want to cause you any inconvenience here, please don't get that wrong: I appreciate your work and wanna thank you for sharing Yasumi! 

But I have to provide a bugfix after such a long time and I can't wait any longer. I don't want to maintain a fork, so I wanted to reach out one more time to ask, if you could share are plans. Is there any chance this gets incorporated anytime soon in a new release?
Is there anything I can do to push this forward?
Shall I recreate the PR from latest `develop` branch? The PR content doesn't look correct to me.

kevinpapst:
Just today:

```
[2024-12-11 10:11:44] [request] CRITICAL: Uncaught PHP Exception DateInvalidTimeZoneException: "DateTimeZone::__construct(): Unknown or bad timezone (Australia/ACT)" at DateTimeZoneFactory.php line 31 {"exception":"[object] (DateInvalidTimeZoneException(code: 0): DateTimeZone::__construct(): Unknown or bad timezone (Australia/ACT) at /www/htdocs/xxxxx/kimai/vendor/azuyalabs/yasumi/src/Yasumi/Provider/DateTimeZoneFactory.php:31)"} []
```

stelgenhof:
Hi @kevinpapst Let me have a look this weekend to make a release.

stelgenhof:
@kevinpapst Sorry to hear that. Unfortunately I haven't been able to work on it as much as I wanted. 
Having a quick look at your PR, it looks a bit odd. Besides some conflicts, it has over 1700 changed files. I assume your PR is not supposed to touch all source files.

kevinpapst:
As I said before: its open-source. I am not expecting anything. I am just sharing my experience. 

> Having a quick look at your PR, it looks a bit odd. Besides some conflicts, it has over 1700 changed files. I assume your PR is not supposed to touch all source files.

Which you force pushed into this branch 🤷 as I said before:

> Unfortunately the maintainer merged a whole bunch of unrelated stuff into this PR

PR #379, opened 2025-07-07
Remove Whitsunday (pentecost) from Austria

felixfrey:
It's not a holiday:
[https://en.wikipedia.org/wiki/Public_holidays_in_Austria](https://en.wikipedia.org/wiki/Public_holidays_in_Austria)
[https://de.wikipedia.org/wiki/Feiertage_in_%C3%96sterreich](https://de.wikipedia.org/wiki/Feiertage_in_%C3%96sterreich)

### DIFF
--- a/src/Yasumi/Provider/Austria.php
+++ b/src/Yasumi/Provider/Austria.php
@@ -55,7 +55,6 @@ class Austria extends AbstractProvider
         $this->addHoliday($this->easterMonday($this->year, $this->timezone, $this->locale));
         $this->addHoliday($this->internationalWorkersDay($this->year, $this->timezone, $this->locale));
         $this->addHoliday($this->ascensionDay($this->year, $this->timezone, $this->locale));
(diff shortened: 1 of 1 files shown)


stelgenhof:
@felixfrey Indeed Whitsunday is not an official public holiday in Austria, but is a major religious observance (As mentioned here https://publicholidays.at/pentecost/ and here https://www.timeanddate.com/holidays/austria/pentecost-sunday).

In Yasumi, Whitsunday is registered for Austria, but not classified as an official holiday. It is classified as "Observance", as you can see in the documentation https://www.yasumi.dev/providers/at.html#holidays and in the source code here: https://github.com/azuyalabs/yasumi/blob/c747552d74e36643ed23a784106289c7b5cfcb88/src/Yasumi/Provider/Austria.php#L58. 

So I don't think it needs to be removed.

felixfrey:
@stelgenhof You are right. I missed the part of the documentation where it explains the filtering methods for holidays. That solves my use case and I hope the filter by type does not break anything for other countries in my code.

PR #393, opened 2025-12-22
add NL equivalent days 2026 till 2028

Stollie:
There are some other holidays I would like to add. 
https://wetten.overheid.nl/BWBR0051300/2025-07-22

These days are needed when for example calculating the number of days some has to pay their bill or days you cannot contact some about their open debt;  https://nvi.nl/nieuws/de-wki-en-de-algemene-termijnenwet


stelgenhof (approved):
Looking good! Thanks again for your contribution.

PR #410, opened 2026-04-13
feat(Australia): refresh holiday rules and rename Queen's Birthday to Monarch's Birthday

aaronflorey:
## Summary
 - update Australian state and territory providers to reflect current holiday rules and timezone handling
 - rename Queen's Birthday references to Monarch's Birthday and add the new translation data
 - add and update coverage for regional holiday behavior, including Anzac Day Monday observance and Easter Sunday/Saturday cases where applicable
 
 ## Testing
 - existing Australia provider tests updated to cover the revised holiday definitions
 - new tests added for state- and region-specific edge cases across affected providers

### DIFF
--- a/src/Yasumi/Provider/Australia.php
+++ b/src/Yasumi/Provider/Australia.php
@@ -35,6 +35,8 @@ class Australia extends AbstractProvider
      */
     public const ID = 'AU';
 
+    protected const MONARCHY_CHANGE_YEAR = 2023;
+
     public string $timezone = 'Australia/Melbourne';
 
     /**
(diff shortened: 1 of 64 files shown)


stelgenhof (changes requested):
@aaronflorey Thank you so much! I have reviewed your PR and it is looking very solid. The only few (non-critical) observations are these

- Consider adding a deprecated alias (`queensBirthday` → `monarchsBirthday`) to ease migration for downstream consumers
- Extract year thresholds (2023, 2024, etc.) into constants for easier maintenance

e.g.:
```
    private const MONARCHY_CHANGE_YEAR = 2023;
    private const SA_EASTER_SUNDAY_YEAR = 2024;
```

- Add specific test for 2022–2023 transition to verify both holiday names appear correctly

stelgenhof (changes requested):
Thanks for the last improvements! All looking very good. There is one small issue left (DCO): see https://github.com/azuyalabs/yasumi/pull/410/checks. Once that is taken care of, it can be merged

stelgenhof (approved):

PR #421, opened 2026-07-18
fix: throw HolidayNotFoundException for unknown holiday keys

zigzagdev:
### What I have done

- `whenIs()` and `whatWeekDayIs()` only validated that the given holiday key was non-blank, not that it actually exists for the current provider/year.                                                                                                   
- Calling `whatWeekDayIs()` with a valid-looking but undefined key (e.g. a holiday that isn't defined for that particular year, or a typo) threw an uncaught `Error: Call to a member function format() on null`.  `whenIs()` silently returned `''` with an "Undefined array key" warning instead of failing clearly.                                                                                                  
- Both methods now go through a shared `getHolidayOrFail()` lookup and throw a new `HolidayNotFoundException` (extends `\InvalidArgumentException`) when the key doesn't resolve to a holiday — consistent with the null-safe contract `getHoliday()` already has.

### Add Test Plans

- Added `testWhenIsWithNonExistentKey` / `testWhatWeekDayIsWithNonExistentKey` to `tests/Base/YasumiTest.php`, using Kenya's `moiDay` (only defined 1989-2009 and 2018-2019) as a non-existent key for a given year.
- Ran the full suite: 18071 tests / 56694 assertions, all green except one pre-existing, unrelated failure  (`Germany\PentecostTest`, which needs the `calendar` PHP extension and is untouched by this change).

### Test Results

```bash

root@be2f1dc89ff6:/app# vendor/bin/phpunit --filter NonExistentKey 
PHPUnit 11.5.56 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.32
Configuration: /app/phpunit.xml.dist

..                                                                  2 / 2 (100%)

Time: 00:00.637, Memory: 142.00 MB

OK (2 tests, 2 assertions)
root@be2f1dc89ff6:/app# 

```


stelgenhof (approved):
Looking good and indeed a very helpful fix!  

Extending the InvalidArgumentException is acceptable as that is how it is done with other exceptions in Yasumi, however I do prefer more precise exception semantics like a maybe a custom NotFound Exception; callers could then catch NotFound exceptions separately from invalid-argument errors. But that is probably something for a next/future release.

